PR #22, opened 2024-01-09
update build pipeline

gzukel:


.github/workflows/build.yml
```diff
@@ -2,10 +2,18 @@ name: "SQS UI | Build Docker Image"
...
+        required: false
+        default: ''
   push:
-    branches: [ "main", "v*.*.*" ]
-  pull_request:
-    branches: [ "main", "v*.*.*" ]
+    tags:
+      - sqs-v[0-9]+.[0-9]+.[0-9]+
+    branches:
+      - "v[0-9]**"
```
p0mvn, line 14: I think that the old one is correct but not the new one. E.g. `v21.x` does not work

.github/workflows/build.yml
```diff
@@ -2,10 +2,18 @@ name: "SQS UI | Build Docker Image"
...
+      docker_tag:
+        description: 'Docker tag to build'
+        required: false
+        default: ''
   push:
-    branches: [ "main", "v*.*.*" ]
-  pull_request:
-    branches: [ "main", "v*.*.*" ]
+    tags:
+      - sqs-v[0-9]+.[0-9]+.[0-9]+
```
p0mvn, line 12: Let's do regular semantic version tags of the format `vx.x.x`

p0mvn (changes requested):

.github/workflows/build.yml
```diff
@@ -2,26 +2,37 @@ name: "SQS UI | Build Docker Image"
...
+        default: ''
   push:
-    branches: [ "main", "v*.*.*" ]
-  pull_request:
-    branches: [ "main", "v*.*.*" ]
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+    branches:
+      - "v[1-9][0-9]?\.[a-z]||[0-9]"
+      - "main"
```
p0mvn, line 15:
```suggestion
      - "osmosis-main"
```

p0mvn (approved):

PR #146, opened 2024-03-20
ci: update steps to latest version and refactor workflows

niccoloraspa:
This PR:

- Updates all steps to the latest versions
- Improves formatting 
- Reads `GO_VERSION` directly from `go.mod`

### DIFF
--- a/.github/workflows/assetlist_test.yml
+++ b/.github/workflows/assetlist_test.yml
@@ -10,19 +10,24 @@ on:
       - "v[0-9]**"
 
 jobs:
-  build:
+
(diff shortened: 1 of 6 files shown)


niccoloraspa:
Is it expected for the routes and prices test to fail?

p0mvn:
> Is it expected for the routes and prices test to fail?

Yes, once we maximize support of prices and routes, we will select a specific number to control for in test. For now, the CI is showing that the support is not maximized

p0mvn (approved):

PR #150, opened 2024-03-26
chore: add v24.x backport to mergify

PaddyMc:
### Description

We are preparing for the v24 deployment of SQS and osmosis-main, we need to have both SQS and osmosis updated to `v24`, we need to do SQS first then update the osmosis node for our CI to be happy.


p0mvn (approved):

p0mvn:
Thank you

PR #162, opened 2024-04-10
Speedup Precompute Order of Magnitude

ValarDragon:
This follows my comments in the Osmosis PR of making it be based off of bitlength. 

OLD:
```
BenchmarkGetPrecomputeOrderOfMagnitude-12    	 5726194	       214.0 ns/op	      96 B/op	       6 allocs/op
```
NEW:
```
BenchmarkGetPrecomputeOrderOfMagnitude-12    	13453051	        78.72 ns/op	      48 B/op	       3 allocs/op
```

### DIFF
--- a/router/usecase/precompute.go
+++ b/router/usecase/precompute.go
@@ -1,53 +1,103 @@
 package usecase
 
-import "github.com/osmosis-labs/osmosis/osmomath"
-
-var (
-	TenE9 = osmomath.NewInt(1_000_000_000)
-	TenE8 = osmomath.NewInt(100_000_000)
-	TenE7 = osmomath.NewInt(10_000_000)
-	TenE6 = osmomath.NewInt(1_000_000)
-	TenE5 = osmomath.NewInt(100_000)
-	TenE4 = osmomath.NewInt(10_000)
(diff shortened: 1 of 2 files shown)


router/usecase/precompute.go
```diff
@@ -1,53 +1,75 @@
...
+			curIndex++
+			bitLenToOrderOfMagnitude = append(bitLenToOrderOfMagnitude, orderLookup{orderOfMagnitude: curIndex, cmpValue: &cmpTen})
+		}
+
+		// Increment the current bit length
+		curBitLen++
+	}
+
+	// Set maxLookupValue to 100 times the next power of ten after maxLookupPowTen
+	maxLookupValue = nextPowTen.QuoRaw(100)
```
p0mvn, line 53: Was this intended to be `Mul` or is `Quo` correct?

p0mvn: I understand why it is not `Mul`, still groking the `100` part

p0mvn: Got it!

p0mvn (approved):
Great work on improving this! Enjoyed the review

PR #167, opened 2024-04-12
Add an LRU cache for tick to sqrt price

ValarDragon:
In CPU profiles we see Tick To Sqrt Price is costly, so here we just cache that call entirely. Along with https://github.com/osmosis-labs/osmosis/pull/8014 I expect this to eliminate most of the current per-call overhead of CL routings. 

This handles https://github.com/osmosis-labs/osmosis/issues/8015

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,8 @@ require (
 	google.golang.org/grpc v1.61.1
 )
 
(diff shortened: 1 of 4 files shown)


p0mvn (approved):
LGTM, thank you!

PR #172, opened 2024-04-14
feat: pricing worker at ingest time

p0mvn:
Splitting pricing logic from the intial prototype with TVL:
https://github.com/osmosis-labs/sqs/pull/159


p0mvn:
complete here:
https://github.com/osmosis-labs/sqs/pull/181

PR #175, opened 2024-04-15
refactor: router refactor, sort pools at ingest but filter when quoting

p0mvn:
This continue the refactor necessary to support pricing at the time of ingest.

In particular it:
- Makes pool sorting happen at the time of ingest but filtering at the time of quote as given by the router configuration. Filternig ~1500 pools should not be expensive but it reduces code complexity and data fragmentation
- As an outcome of that, we will be able to use 3 different routing configurations:
1. Quoting - high min liquidity parameter, higher max routes and max pools per route to find more routes for better quotes
2. Pricing if cache miss by client - lower min liquidity parameter to max out the support of prices. Lower max routes and max pools per route since it has been found to be unnecessary in testing
3. Pricing when precomputing prices at ingest time. 0 min liquidity to maximize the support of all tokens with the desired quote asset


p0mvn:
Remaining steps:
- Fix tests
- Review code
- Load test
- Profile

ingest/usecase/ingest_usecase.go
```diff
@@ -94,6 +93,18 @@ func (p *ingestUseCase) ProcessBlockData(ctx context.Context, height uint64, tak
 	return nil
 }
 
+// sortAndStorePools sorts the pools and stores them in the router.
+// TODO: instead of resorting all pools every block, we should put the updated pools in the correct position
```
p0mvn, line 97: Link

PR #178, opened 2024-04-16
Switch rating to a float

ValarDragon:
This is definitely a pre-mature optimization so we can just ignore this for now. Switched rating to being a float rather than a big Int. 

Floats have 53 bits of precision, so will easily handle precision for pool's TVL's being a billion times (2^30) cheaper than total network TVL.


p0mvn (approved):
LGTM

PR #191, opened 2024-05-02
fix: pools with low liquidity being included in routes

PaddyMc:
### Description

Users have reported getting high price impact for certain routes, this is because when sqs starts we calulate the prices for tokens, but we actually cache the routes, so when we query the routes and there is a cache hit, the incorrect routes will be returned to the end user.

This issue is much more apparent for tokens with many low liquidity pools.

The fix just makes sure that if there is no minimum liquidity value we compute prices without caching.
e.g
https://github.com/osmosis-labs/sqs/blob/v24.x/tokens/usecase/pricing/worker/pricing_worker.go#L92 

### How to test

Run this multiple time to see the price impact:
- `curl https://sqsprod-eu.osmosis.zone/router/quote\?tokenIn\=100000000ibc%2F498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4\&tokenOutDenom\=uion | jq`

then run this branch:
- `curl http://localhost:9092/router/quote\?tokenIn\=100000000ibc%2F498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4\&tokenOutDenom\=uion | jq`

### DIFF
--- a/router/usecase/router_usecase.go
+++ b/router/usecase/router_usecase.go
@@ -143,11 +143,25 @@ func (r *routerUseCaseImpl) GetOptimalQuote(ctx context.Context, tokenIn sdk.Coi
 		rankedRoutes        []route.RouteImpl
 	)
 
-	// If cached ranked routes are not present, compute and rank routes by direct quote
-	if len(candidateRankedRoutes.Routes) == 0 {
-		// Filter pools by minimum liquidity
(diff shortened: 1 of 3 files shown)


cryptomatictrader (approved):
Looks good. To be discussed / implemented is we should consider having separate two route caches - one for pricing api (which needs to include low liq pool), one for route api (which needs to skip low liq pool)

PR #193, opened 2024-05-02
fix: increase min liquidity for routing through pools

PaddyMc:
### Description

- updating the config for min osmo liquidity


cryptomatictrader (approved):
Looks good and necessary. Thanks!

PR #202, opened 2024-05-08
feat: /config-private endpoint, mask otel in /config endpoint

p0mvn:
We block /config endpoint in stage and production because it contains sensitive information

However, it is useful to have it.

As a solution, we expose a separate `/config-private` endpoint with the sensitive config that we can block in LB.

For `/config`, we overwrite the sensitive information, allowing us to unblock it.

### DIFF
--- a/system/delivery/http/system_http_handler.go
+++ b/system/delivery/http/system_http_handler.go
@@ -40,7 +40,7 @@ type JsonResponse struct {
 	} `json:"result"`
 }
 
-// ConfigResponse defines the response for the /config-private endpoint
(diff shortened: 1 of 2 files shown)


system/delivery/http/system_http_handler.go
```diff
@@ -40,6 +40,11 @@ type JsonResponse struct {
 	} `json:"result"`
 }
 
+// ConfigResponse defines the response for the /config-private endpoint
```
p0mvn:
```suggestion
// ConfigPrivateResponse defines the response for the /config-private endpoint
```

PaddyMc (approved):

PR #207, opened 2024-05-08
e2e: test for /tokens/metadata and router/routes

p0mvn:
Progress towards: https://linear.app/osmosis/issue/STABI-140/[sqs-automated-tests-in-ci][candidate-routes]

This PR introduces

`/router/routes and `/tokens/metadata` tests against staging.

It also sets up the foundational infrastructure to allow us to scale this test more towards other environments in the future.

See comments of each test for details about what's being tested and how

## Setup

```bash
make e2e-setup-venv

make e2e-source-venv

make e2e-install-requirements
```

## Running

```bash
# Runs the testsuite in verbose mode
make e2e-run-dev
```

## Next Steps
- Run candidate route test against every supported token and USDC
- Using shared setup infrastructure, run candidate routes tests between combinations of all tokens within the following categories:
  * 5 tokens with high liquidity
  * 5 tokens with low liquidity
  * 5 tokens with high volume
  * 5 tokens with low volume
  * Choose 1 token for which we know, there is a transmuter USDC/USDC.axl, WBTC/WBTC.axl
  * Choose 1 token for which we know, there is a PCL pool - ASTRO


tests/sqs_service.py
```diff
@@ -0,0 +1,64 @@
+import requests
+
+SQS_STAGE = "https://sqs.stage.osmosis.zone"
```
PaddyMc, line 3: :heart:

tests/test_candidate_routes.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        
+        sqs_service = SQSService(environment_url)
+
+        start_time = time.time()
+        response = sqs_service.get_candidate_routes(token_in, token_out)
+        elapsed_time_ms = (time.time() - start_time) * 1000
+
+        assert response.status_code == 200, f"Error: {response.text}"
+        assert expected_latency_upper_bound_ms > elapsed_time_ms, f"Error: latency {elapsed_time_ms} exceeded {expected_latency_upper_bound_ms} ms, token in {token_in} and token out {token_out}"
```
PaddyMc, line 56: hmmm these latency tests could be unwieldy / hard to predict e.g

```
E       AssertionError: Error: latency 342.13829040527344 exceeded 300 ms, token in uosmo and token out ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858
E       assert 300 > 342.13829040527344

tests/test_candidate_routes.py:56: AssertionError
```
||

```
E       AssertionError: Error: latency 363.1315231323242 exceeded 300 ms, token in ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858 and token out uosmo
E       assert 300 > 363.1315231323242
```

Is there some way to improve these? Maybe upper and lower bounds, or something that wouldn't be flaky in e2e, can't really think of something rn, but thought I'd highlight here

PaddyMc (commented):
Awesome work here! Super happy to see some e2e tests in SQS, had one opinion to discuss in the review, but happy besides that 1 comment

tests/test_candidate_routes.py
PaddyMc: Also happy to merge and run in the CI to see if this is even an issue :thinking: 

but we should do that ASAP if that's the case

cryptomatictrader: 300ms makes sense in production but build time may vary a lot depending on how busy the box is at that time.

cryptomatictrader (commented):
Looks good to me. Great work and thanks!

tests/test_candidate_routes.py
p0mvn: Yeah, that's a valid concern.

We will need to select a number that is high enough to not be flaky in CI while also catching regressions if any occur.

300ms was chosen as the right number for my local environment.

Once this is added to CI, I will work on increasing this to ~1s

There is no world where we would expect a query to take that long but if there is some obvious regression, this sanity check would be able to catch it quickly.

PR #209, opened 2024-05-09
STABI-128 - Support Coingecko Prices

cryptomatictrader:
Added Coingecko as the fallback support in /tokens/prices endpoint. In case the chain pricing source is unavailable for any reason, SQS will fallback to using Coingecko as the pricing source.

There is a new parameter in /token/prices called "pricingSource". Possible values are 0 and 1. Default is 0 which represents the chain pricing source and 1 represents the Coingecko pricing source.


tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/osmosis-labs/sqs/domain"
+	"github.com/osmosis-labs/sqs/domain/cache"
+	"github.com/osmosis-labs/sqs/domain/mvc"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	cacheHitsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "sqs_pricing_cache_hits_total",
```
p0mvn: Let's name this as `sqs_pricing_coingecko_cache_hits_total` to not overlap with the chain hits

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,132 @@
...
+			Help: "Total number of pricing cache hits",
+		},
+		[]string{"base", "quote"},
+	)
+	cacheMissesCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "sqs_pricing_cache_misses_total",
```
p0mvn: Similar suggestion

tokens/usecase/tokens_usecase.go
```diff
@@ -246,8 +253,18 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
...
+			var err error
+			price, err = pricingStrategy.GetPrice(ctx, baseDenom, quoteDenom, pricingOptions...)
+			if err != nil { // Check if we should fallback to another pricing source
+				fallbackSourceType := pricingStrategy.ShouldFallback(quoteDenom)
+				if fallbackSourceType != domain.NoneSourceType {
+					fallbackPricingStrategy, ok := t.pricingStrategyMap[fallbackSourceType]
+					if ok {
+						price, err = fallbackPricingStrategy.GetPrice(ctx, baseDenom, quoteDenom, pricingOptions...)
+					}
+				}
```
p0mvn: Could we add a metric or a log here so that we can track which pairs require falling back?

tokens/usecase/tokens_usecase.go
```diff
@@ -371,3 +389,22 @@ func (t *tokensUseCase) IsValidChainDenom(chainDenom string) bool {
...
 }
+
+// IsValidPricingSource implements mvc.TokensUsecase.
+func (t *tokensUseCase) IsValidPricingSource(pricingSource int) bool {
+	switch domain.PricingSourceType(pricingSource) {
+	case domain.ChainPricingSourceType, domain.CoinGeckoPricingSourceType:
+		return true
+	default:
+		return false
+	}
```
p0mvn: nit:

```suggestion
    pricingSourceType := domain.PricingSourceType(pricingSource)
    return  pricingSourceType == domain.ChainPricingSourceType || pricingSourceType == domain.CoinGeckoPricingSourceType
```

tokens/usecase/tokens_usecase.go
```diff
@@ -371,3 +389,22 @@ func (t *tokensUseCase) IsValidChainDenom(chainDenom string) bool {
...
+}
+
+// Get Coingecko ID by chain denom
```
p0mvn: nit: for style consistency
```suggestion
// GetCoingeckoIdByChainDenom gets Coingecko ID by chain denom
```

tokens/usecase/tokens_usecase.go
p0mvn: Let's try to cover this with a test case. We can do so by creating a mock of `PricingSource` and setting up the mock as to force fallback.

See example of how these mocks are used:
https://github.com/osmosis-labs/sqs/tree/v24.x/domain/mocks

We can do this similarly

domain/pricing.go
```diff
@@ -32,6 +33,9 @@ type PricingSource interface {
 	// InitializeCache initialize the cache for the pricing source to a given value.
 	// Panics if cache is already set.
 	InitializeCache(*cache.Cache)
+
+	// ShouldFallback determins what pricing source should be fallen back to in case this pricing source fails
+	ShouldFallback(quoteDenom string) PricingSourceType
```
p0mvn: nit: would it make sense naming this as `GetFallBackStrategy` for readability? I would expect something that starts with `Should` to output a boolean

tokens/usecase/pricing/chain/pricing_chain.go
```diff
@@ -255,3 +255,12 @@ func (c *chainPricing) computePrice(ctx context.Context, baseDenom string, quote
 func (c *chainPricing) InitializeCache(cache *cache.Cache) {
 	c.cache = cache
 }
+
+// Check this strategy allow falling back to another pricing source
```
p0mvn: Recommending for consistency. Whenever we define a method spec on an interface, let's aim to refer to interface directly to avoid fragmenting the context and making it go out-of-date or sync
```suggestion
// ShouldFallback implements pricing.PricingSource
```

domain/mvc/tokens.go
```diff
@@ -40,4 +40,10 @@ type TokensUsecase interface {
 	RegisterPricingStrategy(source domain.PricingSourceType, strategy domain.PricingSource)
 
 	IsValidChainDenom(chainDenom string) bool
+
+	// Check if the pricing source is a valid one
```
p0mvn: nit: I know we have some specs missed above but would still be good to continue working towards maximizing consistency
```suggestion
	// IsValidPricingSource check if the pricing source is a valid one
```

domain/mvc/tokens.go
```diff
@@ -40,4 +40,10 @@ type TokensUsecase interface {
...
+	IsValidPricingSource(pricingSource int) bool
+
+	// Get the Coingecko ID by chain denom
```
p0mvn: ditto:
```suggestion
	// GetCoingeckoIdByChainDenom get the Coingecko ID by chain denom
```

domain/tokens.go
```diff
@@ -7,5 +7,6 @@ type Token struct {
 	// Precision is the precision of the token.
 	Precision int `json:"precision"`
 	// IsUnlisted is true if the token is unlisted.
-	IsUnlisted bool `json:"is_unlisted"`
+	IsUnlisted  bool   `json:"is_unlisted"`
+	CoingeckoID string `json:"coingecko_id"`
```
p0mvn:
```suggestion
        // CoingeckoID the internal ID of this token used by CoinGecko
	CoingeckoID string `json:"coingecko_id"`
```

tokens/delivery/http/tokens_delivery.go
```diff
@@ -123,11 +123,15 @@ func (a *TokensHandler) GetMetadata(c echo.Context) (err error) {
 }
 
 // @Summary Get prices
-// @Description Given a list of base denominations, returns the spot price with a system-configured quote denomination.
+// @Description Given a list of base denominations, this endpoint returns the spot price with a system-configured quote denomination.
+// If the pricing source is set to "chain" (0), it will first check the **chain** pricing cache for the price quote. If it exists, it will return it. Otherwise, it will compute the pricing using block data received from the Osmosis node and store it in the cache with an expiration time specified in the config.json file.
+// If the pricing source is set to "coingecko" (1), a similar pattern applies. It will look for the price quote in the **coingecko** pricing cache. If it exists, it will return it. Otherwise, it will fetch the price from the Coingecko API endpoint and store it in the cache with an expiration time specified in the config.json file.
+// If the token price is not available in the chain pricing source, it will fallback to the Coingecko pricing source if the quote denomination (human or chain) is USDC.
```
p0mvn: This is not quite accurate for chain pricing.

With chain pricing and system-configured quote denomination, we pre-compute prices in the background by detecting tokens that were updated within a block.

As a result, we never expire their TTL.

Since we do not expose the ability for specifying quote denoms, this is the only possible behavior today

tokens/delivery/http/tokens_delivery.go
```diff
@@ -163,11 +171,18 @@ func (a *TokensHandler) GetPrices(c echo.Context) (err error) {
 		}
 	}
 
-	prices, err := a.TUsecase.GetPrices(ctx, baseDenoms, []string{defaultQuoteChainDenom}, domain.ChainPricingSourceType)
+	// Check if pricingSource param is valid
+	pricingSourceInt, err := strconv.Atoi(c.QueryParam("pricingSource"))
```
p0mvn: Could we do stricter validation here?

For example, if a user tries to provide a type but misconfigures due to a typo, we should fail the request.

So I see the following cases:
- "" -> chain source
- 0 -> chain source
- 1 -> coingecko source
- anything else -> error

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,132 @@
...
+		cache:         cache.New(),
+		cacheExpiryNs: time.Duration(config.CacheExpiryMs) * time.Millisecond,
+		quoteCurrency: config.CoingeckoQuoteCurrency,
+		coingeckoUrl:  config.CoingeckoUrl,
+	}
+}
+
+// GetPrice implements pricing.PricingStrategy.
+// quoteDenom is ignored as it uses always coingecko-quote-currency in config.json
+func (c *coingeckoPricing) GetPrice(ctx context.Context, baseDenom string, quoteDenom string, opts ...domain.PricingOption) (osmomath.BigDec, error) {
```
p0mvn, line 90: Could we add a functional style test into CI that is controlled by an environment variable to find all listed tokens as defined by [mainnetState.Tokens](https://github.com/osmosis-labs/sqs/blob/db614d1d38a3d02661a41312866557748ff89a3c/tokens/usecase/pricing/worker/pricing_worker_test.go#L161) that are unsupported in coingecko pricing?

Please see the following example:
https://github.com/osmosis-labs/sqs/blob/db614d1d38a3d02661a41312866557748ff89a3c/tokens/usecase/pricing/worker/pricing_worker_test.go#L146-L150

tokens/usecase/tokens_usecase.go
```diff
@@ -246,8 +253,18 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
 		wg.Add(1)
 		go func(baseDenom, quoteDenom string) {
 			defer wg.Done()
-
-			price, err := pricingStrategy.GetPrice(ctx, baseDenom, quoteDenom, pricingOptions...)
+			var price osmomath.BigDec
+			var err error
+			price, err = pricingStrategy.GetPrice(ctx, baseDenom, quoteDenom, pricingOptions...)
```
p0mvn, line 249: We should also add a unit test for `tokensUseCase.GetPrices(...)` where we specify coingecko pricing option (not fallback but directly) and confirm the right pricing source being called via mocks

p0mvn (commented):
Nice work!

The high-level logic LGTM.

The core of my recommendations is centered on:
- Suggesting test cases
- Stylistic suggestions for consistency

We should also lint the PR with:
```
golangci-lint run --fix
```

tokens/delivery/http/tokens_delivery.go
p0mvn: https://github.com/osmosis-labs/sqs/blob/de34d172f95b221217967799f233c52181cfa07e/README.md#pricing

Added some documentation on this ^

In this PR: https://github.com/osmosis-labs/sqs/pull/214

jonator:
@cryptomatictrader this is great. This means we can remove our own fallback for calculating prices from pool routes on our edge network?

https://github.com/osmosis-labs/osmosis-frontend/blob/028b374f7162952c8285e3d6a507847c6f42819b/packages/server/src/queries/complex/assets/price/providers/sidecar.ts#L30

cryptomatictrader:
@p0mvn , I've updated the ``tokens.json`` by running ``make sqs-update-mainnet-state`` and checked in the ``tokens.json`` only but still it breaks two existing test cases. I will talk to you and I think it's an easy fix.

@jonator , yes! What's the best way to test that out in dev before we push to stage? Maybe we push to dev.osmosis.zone and do some pair?

app/sqs_config.go
```diff
@@ -43,8 +43,10 @@ var DefaultConfig = domain.Config{
...
 		DefaultQuoteHumanDenom: "usdc",
 
-		MaxPoolsPerRoute: 4,
-		MaxRoutes:        5,
-		MinOSMOLiquidity: 50,
+		MaxPoolsPerRoute:       4,
+		MaxRoutes:              5,
+		MinOSMOLiquidity:       50,
+		CoingeckoUrl:           "https://prices.osmosis.zone/api/v3/simple/price",
+		CoingeckoQuoteCurrency: "usd",
```
p0mvn, line 50: Prior to merge, let's make sure that our config templating and infrastructure repo is updated to be able to deploy the new values correctly.

We will need some modifications in the following areas:
- https://github.com/osmosis-labs/infrastructure/blob/e78fbf64adb95a9fcbc421b76b6afeb68d1d1200/environments/sqs-osmosis-zone/ansible/roles/prepare_local_folder/templates/sqs/config.json.j2
- https://github.com/osmosis-labs/infrastructure/blob/e78fbf64adb95a9fcbc421b76b6afeb68d1d1200/environments/sqs-osmosis-zone/environments/stage/ansible/vars/sqs.yaml

Once ready for production, we will also need to update this:
- https://github.com/osmosis-labs/infrastructure/blob/e78fbf64adb95a9fcbc421b76b6afeb68d1d1200/environments/sqs-osmosis-zone/environments/prod/ansible/vars/sqs.yaml

p0mvn, line 50: Could we also update `config_testnet.json` with the new values?

domain/pricing.go
```diff
@@ -19,6 +19,7 @@ const (
 	// CoinGeckoPricingSourceType defines the pricing source
 	// that calls CoinGecko API.
 	CoinGeckoPricingSourceType
+	NoneSourceType
```
p0mvn: Should we instead make `NoneSourceType` as the first element or explicitly set it to -1?

The problem with the current approach is that if we keep adding more types in the future and placing them before `NoneSourceType` in the enum, the value of `NoneSourceType` will keep increasing with every addition, potentially leading to breaking changes.

router/usecase/routertesting/suite.go
```diff
@@ -344,6 +346,12 @@ func (s *RouterTestHelper) SetupRouterAndPoolsUsecase(mainnetState MockMainnetSt
 
 	tokensUsecase.RegisterPricingStrategy(domain.ChainPricingSourceType, pricingSource)
 
+	// Set up Coingecko pricing strategy
+	options.PricingConfig.DefaultSource = domain.CoinGeckoPricingSourceType
+	coingeckoPricingSource, err := pricing.NewPricingStrategy(options.PricingConfig, tokensUsecase, routerUsecase)
```
p0mvn: Should we instead be registering a mock coingecko pricing strategy?

Otherwise, our unit tests may start making network requests which is not ideal, leading to flakiness and increased latency

tokens/delivery/http/tokens_delivery.go
```diff
@@ -123,11 +123,16 @@ func (a *TokensHandler) GetMetadata(c echo.Context) (err error) {
 }
 
 // @Summary Get prices
-// @Description Given a list of base denominations, returns the spot price with a system-configured quote denomination.
+// @Description Given a list of base denominations, this endpoint returns the spot price with a system-configured quote denomination.
+// If the pricing source is set to "chain" (0), it will first check the **chain** pricing cache for the price quote. If it exists, it will return it. Otherwise, it will compute the pricing on-demand if the quote is non-usdc.
+// If the pricing source is set to "coingecko" (1), it will look for the price quote in the **coingecko** pricing cache. If it exists, it will return it. Otherwise, it will fetch the price from the Coingecko API endpoint and store it in the cache with an expiration time specified in the config.json file.
+// If the token price is not available from the chain pricing source for any reason, it will fallback to the Coingecko pricing source if the quote denomination (human or chain) is usdc.
+// See also: https://github.com/osmosis-labs/sqs/blob/de34d172f95b221217967799f233c52181cfa07e/README.md#pricing
```
p0mvn, line 130: Nice! Could you please also extend the CoinGecko section documentation here:
https://github.com/osmosis-labs/sqs/blob/v24.x/README.md#coingecko

In a subsequent effort, I propose we move some of this detail under `docs/architecture/pricing.md` to avoid having the README grow out of control.

Instead, we can link specific documents(such as `pricing.md`) from the top level README

tokens/delivery/http/tokens_delivery.go
```diff
@@ -163,11 +172,23 @@ func (a *TokensHandler) GetPrices(c echo.Context) (err error) {
...
+	pricingSourceParam := c.QueryParam("pricingSource")
+	if pricingSourceParam == "" {
+		pricingSource = domain.ChainPricingSourceType
+	} else {
+		pricingSourceInt, err := strconv.Atoi(pricingSourceParam)
+		if err != nil || !a.TUsecase.IsValidPricingSource(pricingSourceInt) {
+			return c.JSON(http.StatusBadRequest, domain.ResponseError{Message: fmt.Sprintf("invalid pricing source: %s", pricingSourceParam)})
+		}
+		pricingSource = domain.PricingSourceType(pricingSourceInt)
+	}
```
p0mvn:
```suggestion
	if pricingSourceParam == "" {
	         // By default, if no source is explicitly provided, we use a chain pricing source. 
		pricingSource = domain.ChainPricingSourceType
	} else {
	        // Otherwise, we attempt to parse the given pricing source.
		pricingSourceInt, err := strconv.Atoi(pricingSourceParam)
		if err != nil || !a.TUsecase.IsValidPricingSource(pricingSourceInt) {
			return c.JSON(http.StatusBadRequest, domain.ResponseError{Message: fmt.Sprintf("invalid pricing source: %s", pricingSourceParam)})
		}
		pricingSource = domain.PricingSourceType(pricingSourceInt)
	}
```

tokens/usecase/pricing/chain/pricing_chain.go
```diff
@@ -253,3 +253,12 @@ func (c *chainPricing) computePrice(ctx context.Context, baseDenom string, quote
...
 }
+
+// GetFallbackStrategy implements pricing.PricingSource
+func (c *chainPricing) GetFallbackStrategy(quoteDenom string) domain.PricingSourceType {
+	if quoteDenom == c.defaultQuoteDenom {
+		return domain.CoinGeckoPricingSourceType
+	} else {
+		return domain.NoneSourceType
+	}
+}
```
p0mvn, line 283: Nice

tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var (
+	defaultRouterConfig  = routertesting.DefaultRouterConfig
+	defaultPricingConfig = routertesting.DefaultPricingConfig
+)
+
+func (suite *CoingeckoPricingTestSuite) BeforeTest(suiteName, testName string) {
+	// Run some code before each test
+}
```
p0mvn: We can ignore this if setup is not needed
```suggestion
```

tokens/usecase/tokens_usecase_test.go
```diff
@@ -135,6 +142,21 @@ func (s *TokensUseCaseTestSuite) TestParseExponents_Testnet() {
 	s.Require().Equal(defaultCosmosExponent, osmoToken.Precision)
 }
 
+func (s *TokensUseCaseTestSuite) TestGetPrices_Coingecko() {
```
p0mvn, line 147: Do you mind adding a short comment summary, specifying what this test aims to validate?

tokens/usecase/tokens_usecase_test.go
```diff
@@ -135,6 +142,21 @@ func (s *TokensUseCaseTestSuite) TestParseExponents_Testnet() {
...
+	// Set up mainnet mock state.
+	mainnetUsecase := s.SetupDefaultRouterAndPoolsUsecase()
+	prices, err := mainnetUsecase.Tokens.GetPrices(context.Background(), routertesting.MainnetDenoms, []string{USDC}, domain.CoinGeckoPricingSourceType)
```
p0mvn, line 150: I would like to note that this test is making network requests to Coingecko. Instead, we should monkey-patch a mock function that would simulate the interaction with Coingecko.

It would be challenging to relay in words what I mean so I did a quick demo here:
https://github.com/osmosis-labs/sqs/pull/218

Please let me know if you have any questions - we can discuss during data sync tomorrow

p0mvn (commented):
Nice work! Had some follow-up comments

p0mvn:
> @p0mvn , I've updated the `tokens.json` by running `make sqs-update-mainnet-state` and checked in the `tokens.json` only but still it breaks two existing test cases. I will talk to you and I think it's an easy fix.
> 
> @jonator , yes! What's the best way to test that out in dev before we push to stage? Maybe we push to dev.osmosis.zone and do some pair?

Sounds great.

For testing/deployment, let's:
- Address any remaining comments and get this approved
- Update stage configs in the infrastructure repository
- Merge the PR
- Manually test https://dev.osmosis.zone/
- Request QA of https://dev.osmosis.zone/ in #eng-qa, specifying the context around what has changed and testing suggestions

Look forward to seeing this change live soon!

cryptomatictrader:
Thanks a lot! All good comments and tips. I'll work on them 👍

app/sqs_config.go
cryptomatictrader: Pull request created https://github.com/osmosis-labs/infrastructure/pull/243

cryptomatictrader: Done

domain/pricing.go
cryptomatictrader: I'll explicitly set `NoneSourceType` to `-1` and place it at the end, so that `ChainPricingSourceType` always stays at `0`, `CoinGeckoPricingSourceType` stays at `1`, and if we have more pricing sources in future, they will be 2, 3, 4...

tokens/delivery/http/tokens_delivery.go
cryptomatictrader: Thanks for the comment suggestion.

tokens/usecase/tokens_usecase_test.go
cryptomatictrader: Done

tokens/delivery/http/tokens_delivery.go
cryptomatictrader: @p0mvn , can you share me the link to pricing.md ? I don't seem to find it in https://github.com/osmosis-labs/docs/tree/main Thanks!

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+		[]string{"base", "quote"},
+	)
+)
+
+// CoingeckoPriceGetterFn is a function type that fetches the price of a token from Coingecko.
+// We monkey-patch this function for testing purposes.
+type CoingeckoPriceGetterFn func(ctx context.Context, baseDenom string, coingeckoId string) (osmomath.BigDec, error)
+
+// DefaultCoingeckoPriceGetter is the default implementation of CoingeckoPriceGetterFn, which invokes GetPriceByCoingeckoId.
```
p0mvn:
```suggestion
// DefaultCoingeckoPriceGetter represents a placeholder for the default implementation of CoingeckoPriceGetterFn, which invokes GetPriceByCoingeckoId method that is defined on coingeckoPricing.
```

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,168 @@
...
+var DefaultCoingeckoPriceGetter CoingeckoPriceGetterFn = nil
+
+// MockCoingeckoPriceGetter is a mock implementation of CoingeckoPriceGetterFn, which returns a fixed price of 1.
+var MockCoingeckoPriceGetter CoingeckoPriceGetterFn = func(ctx context.Context, baseDenom, coingeckoId string) (osmomath.BigDec, error) {
+	if coingeckoId == "" {
+		return osmomath.NewBigDec(0), nil
+	} else {
+		return osmomath.NewBigDec(1), nil
+	}
+}
```
p0mvn: nit: let's move mock to a test file?

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,168 @@
...
+// We monkey-patch this function for testing purposes.
+type CoingeckoPriceGetterFn func(ctx context.Context, baseDenom string, coingeckoId string) (osmomath.BigDec, error)
+
+// DefaultCoingeckoPriceGetter is the default implementation of CoingeckoPriceGetterFn, which invokes GetPriceByCoingeckoId.
+var DefaultCoingeckoPriceGetter CoingeckoPriceGetterFn = nil
+
+// MockCoingeckoPriceGetter is a mock implementation of CoingeckoPriceGetterFn, which returns a fixed price of 1.
+var MockCoingeckoPriceGetter CoingeckoPriceGetterFn = func(ctx context.Context, baseDenom, coingeckoId string) (osmomath.BigDec, error) {
+	if coingeckoId == "" {
+		return osmomath.NewBigDec(0), nil
```
p0mvn: nit: if we don't expect the results to be modified let's aim to define `osmomath` constants in the global space. These types take minor overhead for the initialization of internal buffers.

While negligible individually, these may add meaningful overhead to test times when accumulated

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,168 @@
...
+		// Set the custom price getter function (useful for testing purposes)
+		coingeckoPricing.priceGetterFn = coingeckoPriceGetterFn
+	}
+
+	return coingeckoPricing
+}
+
+// GetPrice implements pricing.PricingStrategy.
+// quoteDenom is ignored as it uses always coingecko-quote-currency in config.json
+func (c *coingeckoPricing) GetPrice(ctx context.Context, baseDenom string, quoteDenom string, opts ...domain.PricingOption) (osmomath.BigDec, error) {
```
p0mvn, line 90: Using the mocking infrastructure that was just integrated, would it be possible to add more table-driven unit-tests that maximize all of the edge case coverage while treating this function as a black box?

For example, covering the fact that `quoteDenom` is ignored via a table-driven unit test case.

I know that we already have a higher-level test in `tokensUseCase` but I think that also having a stricter test at a lower level of abstraction would pay off in the long term

tokens/usecase/pricing/pricing.go
```diff
@@ -7,13 +7,17 @@ import (
...
 )
 
 // NewPricingStrategy is a factory method to create the pricing strategy based on the desired source.
 func NewPricingStrategy(config domain.PricingConfig, tokensUsecase mvc.TokensUsecase, routerUseCase mvc.RouterUsecase) (domain.PricingSource, error) {
 	if config.DefaultSource == domain.ChainPricingSourceType {
 		return chainpricing.New(routerUseCase, tokensUsecase, config), nil
 	}
+	if config.DefaultSource == domain.CoinGeckoPricingSourceType {
+		return coingeckopricing.New(routerUseCase, tokensUsecase, config, coingeckopricing.DefaultCoingeckoPriceGetter), nil
+	}
```
p0mvn: It feels odd that we pass in an entire config while only needing the `DefaultSource` field.

Should we instead pass in a `domain.CoinGeckoPricingSourceType` variable?

tokens/usecase/pricing/worker/pricing_worker_test.go
```diff
@@ -231,7 +231,9 @@ func (s *PricingWorkerTestSuite) TestGetPrices_Chain_FindUnsupportedTokens() {
 	// SHARK - CW pool (unsupported in tests but should be supported on mainnet)
 	// FURY - listed but no pools
 	// FURY.legacy - listed but no pools
-	s.Require().Equal(25, zeroPriceCounter)
+	//
+	// Update on May 13, 2024: 14 unsupported tokens because some tokens have been fallen back to backup pricining source Coingecko
+	s.Require().Equal(14, zeroPriceCounter)
```
p0mvn, line 236: Nice. In the future, we will need to think about making this more dynamic so that when we update internal test state, these assertions would not break.

However, this is a future goal that is outside of scope of this PR - just noting

p0mvn (commented):
Nice job! 💪 

Had a few more recommendations but almost there!

tokens/usecase/pricing/pricing.go
cryptomatictrader: But we still need the `config` when calling the `New` in `chainpricing.New` and `coingeckopricing.New`

tokens/usecase/pricing/worker/pricing_worker_test.go
cryptomatictrader: 👍  Agree. I don't like the magic number there as well

tokens/usecase/pricing/coingecko/pricing_coingecko.go
cryptomatictrader: Hmm, looks like moving it `MockCoingeckoPriceGetter` to `pricing_coingecko_test.go` will break `suite.go` as `suite.go` cannot reference anything defined in `pricing_coingecko_test.go`. Maybe I'm missing something?

tokens/usecase/pricing/coingecko/pricing_coingecko.go
cryptomatictrader: Having table-driven unit test cases make sense. But as you said, there may be some test cases duplication with `tokens_usecase_test.go`, what edge cases are you you thinking about? 

Thanks!

tokens/usecase/pricing/coingecko/pricing_coingecko.go
p0mvn: Could we instead move it into a dedicated `mocks` package in a separate file for consistency with other mocks?

https://github.com/osmosis-labs/sqs/blob/0289f47a375c467588a8564e12d4cf4ea646a7b0/domain/mocks

In the future, we can exclude that package from build with build flags

tokens/usecase/pricing/coingecko/pricing_coingecko.go
p0mvn: It is acceptable to have overlapping unit tests at various levels of abstraction.

Having them defined in such a way would allow us, developers, to identify the exact source of a bug or regression if any arise.

Ideally, we should aim to treat a function as a black box. Given the abstraction, its state, and all possible combinations of inputs and outputs - what are the test cases that make sense to construct? 

In this case, I would add the following cases:
- happy path case (quote denom defined on the struct is used for getting the price and same quote denom is provided as function parameter)
- the quote denom provided as input to the function is ignored and another denom that is defined on the struct is used instead)
- some other non-standard quote denom is set on the struct (e.g. instead of usdc, set usdt) 
- mock error being returned and bubbled up to the client
  * This case is important because we have various assumptions around the system around error handling. In some areas, we silently ignore them to avoid breaking the client and return zero instead. In other cases, we actually bubble up the error. As a result, it is important to keep these invariants strict with test coverage.

tokens/usecase/pricing/pricing.go
p0mvn: You're right. I must have missed something - apologies

tokens/usecase/pricing/coingecko/pricing_coingecko.go
cryptomatictrader: True - it does help identifying where the root cause may be. Let's sync up tomorrow (Wed). I'll show you what I've done so far and make sure I'm on the right track. Thanks!

cryptomatictrader: @p0mvn , I've enhanced and test cases. Feel free to comment it. Thanks!!

tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
```diff
@@ -0,0 +1,117 @@
...
+	mainnetUsecase := s.SetupDefaultRouterAndPoolsUsecase()
+	defaultPricingConfig.DefaultSource = domain.CoinGeckoPricingSourceType
+	coingeckoPricingSource := coingeckopricing.New(mainnetUsecase.Router, mainnetUsecase.Tokens, defaultPricingConfig, mocks.DefaultMockCoingeckoPriceGetter)
+
+	tests := []struct {
+		desc       string
+		baseDenom  string
+		quoteDenom string
+		shouldZero bool
+		shouldErr  bool
```
p0mvn: I think we can have one of these booleans since in the test cases both of these are either true or false

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -0,0 +1,167 @@
...
+		panic(err)
+	}
+	if err := prometheus.Register(cacheMissesCounter); err != nil {
+		panic(err)
+	}
+}
+
+// New creates a new Coingecko pricing source.
+// if coinGeckoPriceGetterFn is nil, it uses the default implementation.
+func New(routerUseCase mvc.RouterUsecase, tokenUseCase mvc.TokensUsecase, config domain.PricingConfig, coingeckoPriceGetterFn CoingeckoPriceGetterFn) domain.PricingSource {
```
p0mvn: `routerUseCase` is unused

tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
```diff
@@ -0,0 +1,117 @@
...
+		{"Test coingecko GetPrice with quote denom as some spaces", ATOM, " ", false, false},
+		{"Test coingecko GetPrice with quote denom as ATOM", ATOM, ETH, true, true},
+		{"Test coingecko GetPrice with invalid base denom", "-DUMMY-", USDC, true, true},
+		{"Test coingecko GetPrice with empty base denom", "", USDC, true, true},
+		{"Test coingecko GetPrice with some spaces as base denom", " ", USDC, true, true},
+	}
+
+	for _, tt := range tests {
+		s.Run(tt.desc, func() {
+			price, err := coingeckoPricingSource.GetPrice(context.Background(), tt.baseDenom, tt.quoteDenom)
```
p0mvn, line 74: Is it possible to also do some assertions on the value of `price`?

For example, on success cases, we could set up the mock in a way where if "ATOM" and "USDC" are given , the returned price is 0.5 but if it is any other denom, it would return the default of 1.

tokens/usecase/pricing/coingecko/pricing_coingecko.go
cryptomatictrader: Good catch!

tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
cryptomatictrader: I propose we have shouldErr, expectedPrice in the struct and not just one field to overload and represent two meaning. Please let me know what you think. Thanks!

tokens/usecase/pricing/coingecko/pricing_coingecko.go
```diff
@@ -89,11 +89,11 @@ func New(routerUseCase mvc.RouterUsecase, tokenUseCase mvc.TokensUsecase, config
 // So quoteDenom has to be nil or usdc or usdt
 func (c *coingeckoPricing) GetPrice(ctx context.Context, baseDenom string, quoteDenom string, opts ...domain.PricingOption) (osmomath.BigDec, error) {
 	if quoteDenom != USDC_DENOM && quoteDenom != USDT_DENOM && strings.TrimSpace(quoteDenom) != "" {
-		return osmomath.BigDec{}, fmt.Errorf("only usdc/usdt denom or nil is allowed for the quote denom param")
+		return osmomath.ZeroBigDec(), fmt.Errorf("only usdc/usdt denom or nil is allowed for the quote denom param")
 	}
 	coingeckoId, err := c.TUsecase.GetCoingeckoIdByChainDenom(baseDenom)
 	if err != nil {
-		return osmomath.BigDec{}, err
+		return osmomath.ZeroBigDec(), err
```
p0mvn: nit: what was before is a more idiomatic way in Go. If we return an error, the value should be initialized. Suggest reverting this for consistency with conventions

p0mvn (approved):
Great! I didn't run the latest test updates but they look working to me!

Nice work on this effort!

tokens/usecase/pricing/coingecko/pricing_coingecko.go
cryptomatictrader: I thought `ZeroBigDec` was intentionally created as a convenient function for us to get a zero `BigDec` and I saw it does `func ZeroBigDec() BigDec { return BigDec{new(big.Int).Set(zeroInt)} }` internally. I will change it to `BigDec{}` as it is more idiomatic. Thanks for all the review efforts!!!!

PR #210, opened 2024-05-09
DATA-33 Migrate to asset list v2

cryptomatictrader:
This pull request address the use of asset list. SQS needs to sunset the use of v1 asset list and move to v2 asset list.


cryptomatictrader:
Will review with Roman and work out/perform the test efforts before we merge.

cryptomatictrader:
Haven't persisted the coingecko id. Will add it in

tokens/usecase/tokens_usecase.go
```diff
@@ -32,24 +32,74 @@ type tokensUseCase struct {
...
+		CoingeckoID string `json:"coingeckoId"`
+		Price       struct {
+			PoolID string `json:"poolId"`
+			Denom  string `json:"denom"`
+		} `json:"price"`
+		Categories   []string `json:"categories"`
+		PegMechanism string   `json:"pegMechanism"`
+		// Found inconsistency in the assest list file
+		// TransferMethods/CounterParty sometimes is an array and sometimes is an object
+		// So commenting out the fields for now
```
p0mvn: I'm supportive of removing all fields that we don't need and skip parsing them. Instead, leave the comment that this is what we are trying to do

tokens/usecase/tokens_usecase.go
```diff
@@ -302,41 +352,11 @@ func GetTokensFromChainRegistry(chainRegistryAssetsFileURL string) (map[string]d
...
-		}
-
+		token.Precision = asset.Decimals
 		token.HumanDenom = asset.Symbol
-		token.IsUnlisted = isUnlisted
-
-		tokensByChainDenom[chainDenom] = token
+		token.IsUnlisted = asset.Disabled
+		token.CoingeckoID = asset.CoingeckoID
+		tokensByChainDenom[asset.CoinMinimalDenom] = token
```
p0mvn: Looks like v2 has a much cleaner format

p0mvn (commented):
Great work! Just wondering - are there any tests that would make sense adding here or do you think the existing ones already provide sufficient coverage?

tokens/usecase/tokens_usecase.go
cryptomatictrader: Removed unused asset list definitions

tokens/usecase/tokens_usecase.go
cryptomatictrader: Please note ``is_unlisted`` maps to ``preview``, not ``disabled``. Fixed.

cryptomatictrader:
I believe the existing unit tests should cover it quite well because not constructing the token object with proper values will break the test cases pretty hard. But I'll also run the e2e test to make sure. Will talk to you soon about running the tests. Thanks!

p0mvn (approved):
Great work!

Recommending the following plan:
- (If not yet done) test this change out on a running dev node and validate a few swap pairs with swagger
- Let's merge this
- Backport to the `v25.x` branch via a backport label
- Create a tag and update it [here](https://github.com/osmosis-labs/infrastructure/blob/3d0297c0d19a5efa087e6e1980f59ad499e042e8/environments/sqs-osmosis-zone/environments/stage/ansible/vars/version.yaml#L7) for staging
- Deploy the change to stage and do a round of manual QA
- Request #eng-qa to do a round of QA in https://dev.osmosis.zone/ to make sure that most pairs continue routing
- On success, modify [this](https://github.com/osmosis-labs/infrastructure/blob/3d0297c0d19a5efa087e6e1980f59ad499e042e8/environments/sqs-osmosis-zone/environments/prod/ansible/vars/version.yaml#L7) tag for prod
- Deploy prod

cryptomatictrader:
@p0mvn thanks for the review.

Just want to make sure I follow the proper process, given that v25.x becomes current.

1. I've merged the latest code from `v25.x` branch into this branch locally, ran some tests using swagger and they are all good. Should I proceed to push the merged changes to `calvin/migrate-to-asset-list-v2`
2. Will create a tag after the push
3. You mentioned I need to update the `version.yaml` for staging and deploy it to stage but ask QA to run a round of QA in https://dev.osmosis.zone. Do both https://dev.osmosis.zone (as well as https://stage.osmosis.zone) use staging 'version.yaml'

domain/tokens.go
```diff
@@ -3,9 +3,10 @@ package domain
...
 	// HumanDenom is the human readable denom.
-	HumanDenom string `json:"human_denom"`
+	HumanDenom string `json:"symbol"`
 	// Precision is the precision of the token.
-	Precision int `json:"precision"`
+	Precision int `json:"decimals"`
 	// IsUnlisted is true if the token is unlisted.
-	IsUnlisted bool `json:"is_unlisted"`
+	IsUnlisted  bool   `json:"preview"`
+	CoingeckoID string `json:"coingeckoId"`
```
p0mvn, line 11: Just realized that this would be a breaking change.

For example, it would break our new integration tests. I don't think there are any other clients so we can simply fix our integration tests.

Could you please confirm that the tests continue to work by running `e2e-run-dev`?

p0mvn (changes requested):

p0mvn:
> @p0mvn thanks for the review.
> 
> Just want to make sure I follow the proper process, given that v25.x becomes current.
> 
> 1. I've merged the latest code from `v25.x` branch into this branch locally, ran some tests using swagger and they are all good. Should I proceed to push the merged changes to `calvin/migrate-to-asset-list-v2`
> 2. Will create a tag after the push
> 3. You mentioned I need to update the `version.yaml` for staging and deploy it to stage but ask QA to run a round of QA in https://dev.osmosis.zone. Do both https://dev.osmosis.zone (as well as https://stage.osmosis.zone) use staging 'version.yaml'


Good questions!

1 and 2. We should not merge `v25.x` branch into this one. Instead, can merge this PR directly to `v24.x` while applying the "A:backport/v25.x" label. On merge to `v24.x`, mergify will auto-create a backport PR to v25.x for us. Then, we can merge that to `v25.x` and create a tag.
3. 
- https://dev.osmosis.zone uses https://sqs.stage.osmosis.zone/swagger/index.html#/default/get-token-metadata
- https://stage.osmosis.zone/ uses sqs.osmosis.zone/

So Dev FE => SQS Stage and Stage FE => SQS Prod

This is so that FE developers have their environment as close to production as possible and so that the risky changes that we test in dev do not affect their environment

domain/tokens.go
p0mvn: https://github.com/osmosis-labs/sqs/blob/5062373fcf18dd144236a9c1ae4311d5be21a1b0/tests/setup.py#L286

cryptomatictrader: Addressed in https://github.com/osmosis-labs/sqs/pull/210/commits/170cd24df7f97a367ab4d842f863806f59e61bb9

cryptomatictrader:
@p0mvn , please note I've changed the base branch to v25.x. e2e-run-dev has passed in my local environment. Will create a tag and update staging after review and merge into v25.x.

.github/mergify.yml
```diff
@@ -54,3 +54,11 @@ pull_request_rules:
...
         branches:
           - v24.x
+  - name: backport patches to v25.x branch
+    conditions:
+      - base=v24.x
+      - label=A:backport/v25.x
+    actions:
+      backport:
+        branches:
+          - v25.x
```
p0mvn, line 64: Ahh that's why it didn't work for me

I saw this present on `v24.x` but did not check `v25.x`

Thanks for adding

CHANGELOG.md
```diff
@@ -43,13 +43,18 @@ Ref: https://keepachangelog.com/en/1.0.0/
...
 that is already correctly scaled when computing the price using the alternative (quote-based) method.
 
 ## v0.17.10
 
 - /config-private endpoint, mask OTEL config in /config endpoint
 
+## v0.17.9
+
+Fixes for pricing cache and min liquidity param
+
```
p0mvn, line 57: Thanks for bringing these over

p0mvn (approved):
🥪

PR #212, opened 2024-05-10
fix: DYDX spot price bug

p0mvn:
Fixes this:
image

The core of the bug was that in edge cases the `TokenOutDenom` is not set by design. As a result, the quote-based spot price method for Astroport PCL pools would fail, falling back to the CW pool query that would return spot price that is already scaled.

That would result in falling back to querying the contract directly but the contract returns already descaled spot price (relative to denom precision by mistake)

### Prod Prices

```

  "ibc/831F0B1BBB1D08A2B75311892876D71565478C532967545476DF4C2D7492E48C": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "2164876635285.779993175913358528062425000000000000"
  },
  "ibc/B8C608CEE08C4F30A15A7955306F2EDAF4A02BB191CABC4185C1A57FD978DA1B": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "0.095202519949110931378352595589401691"
  },
  "uosmo": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "0.871430586109772434016678413411669181"
  }
}
```

This PR prices

```
{
  "ibc/831F0B1BBB1D08A2B75311892876D71565478C532967545476DF4C2D7492E48C": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "2.167893220913947883791043000000000000"
  },
  "ibc/B8C608CEE08C4F30A15A7955306F2EDAF4A02BB191CABC4185C1A57FD978DA1B": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "0.095391145246243553183855132539200743"
  },
  "uosmo": {
    "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4": "0.871361998916329797000000000000000000"
  }
}
```

⏫ Note that all prices are the same but DYDX is fixed

### DIFF
--- a/router/usecase/router_usecase.go
+++ b/router/usecase/router_usecase.go
@@ -705,7 +705,7 @@ func (r *routerUseCaseImpl) GetPoolSpotPrice(ctx context.Context, poolID uint64,
 		return osmomath.BigDec{}, fmt.Errorf("taker fee not found for pool %d, denom in (%s), denom out (%s)", poolID, quoteAsset, baseAsset)
 	}
 
-	spotPrice, err := r.poolsUsecase.GetPoolSpotPrice(ctx, poolID, poolTakerFee, baseAsset, quoteAsset)
+	spotPrice, err := r.poolsUsecase.GetPoolSpotPrice(ctx, poolID, poolTakerFee, quoteAsset, baseAsset)
 	if err != nil {
 		return osmomath.BigDec{}, err
 	}
--- a/sqsdomain/routable_pool.go
+++ b/sqsdomain/routable_pool.go
@@ -33,9 +33,6 @@ type RoutablePool interface {
 	CalculateTokenOutByTokenIn(ctx context.Context, tokenIn sdk.Coin) (sdk.Coin, error)
 	ChargeTakerFeeExactIn(tokenIn sdk.Coin) (tokenInAfterFee sdk.Coin)
 
-	// SetTokenOutDenom sets the token out denom on the routable pool.
-	SetTokenOutDenom(tokenOutDenom string)
-
 	GetTakerFee() osmomath.Dec
 
 	GetSpreadFactor() osmomath.Dec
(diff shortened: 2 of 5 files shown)


router/usecase/router_usecase.go
```diff
@@ -705,7 +705,7 @@ func (r *routerUseCaseImpl) GetPoolSpotPrice(ctx context.Context, poolID uint64,
 		return osmomath.BigDec{}, fmt.Errorf("taker fee not found for pool %d, denom in (%s), denom out (%s)", poolID, quoteAsset, baseAsset)
 	}
 
-	spotPrice, err := r.poolsUsecase.GetPoolSpotPrice(ctx, poolID, poolTakerFee, baseAsset, quoteAsset)
+	spotPrice, err := r.poolsUsecase.GetPoolSpotPrice(ctx, poolID, poolTakerFee, quoteAsset, baseAsset)
```
p0mvn, line 708: Note: This is a drive-by change. We were negating this in `pricing_chain.go` so it did not result in any bug

tokens/usecase/pricing/chain/pricing_chain.go
```diff
@@ -221,8 +221,6 @@ func (c *chainPricing) computePrice(ctx context.Context, baseDenom string, quote
 	if useAlternativeMethod {
 		// Compute on-chain price for 1 unit of base denom and quote denom.
 		chainPrice = osmomath.NewBigDecFromBigInt(tenQuoteCoin.Amount.BigIntMut()).QuoMut(osmomath.NewBigDecFromBigInt(quote.GetAmountOut().BigIntMut()))
-	} else {
-		chainPrice = osmomath.OneBigDec().QuoMut(chainPrice)
```
p0mvn, line 225: Note: This is a drive-by change. This was needed because we were confusing base and quote in `router_usecase.go`

sqsdomain/routable_pool.go
```diff
@@ -33,9 +33,6 @@ type RoutablePool interface {
 	CalculateTokenOutByTokenIn(ctx context.Context, tokenIn sdk.Coin) (sdk.Coin, error)
 	ChargeTakerFeeExactIn(tokenIn sdk.Coin) (tokenInAfterFee sdk.Coin)
 
-	// SetTokenOutDenom sets the token out denom on the routable pool.
-	SetTokenOutDenom(tokenOutDenom string)
```
PaddyMc, line 37: Can we remove this at a later date? Will this break the node, if `sqsdomain` isn't updated?

p0mvn: No, this will not break the node.

It would only break if we changed serialization.

PaddyMc: Great, happy to approve then

PaddyMc (approved):

PR #214, opened 2024-05-10
Clean up chain pricing; add quote-based and spot-price based tests; add docs

p0mvn:
Implement this refactor, tests and docs while debugging the DYDX issue.

While the bug was present at a different layer, this is still an improvement so making the PR.

- Clean up chain pricing by adding an option to recompute prices using a quote-based method for testability
- Improve identifier namings and add code comments
- Add a unit test for validating quote-based and spot-price based methods together
- Add pricing docs to the README


cryptomatictrader (commented):
All look good to me!

p0mvn:
Thanks for review - merging

PR #216, opened 2024-05-11
POC: Flight recording of slow requests (above 500ms)

p0mvn:
Enables flight recording from this article: https://go.dev/blog/execution-traces-2024

It allows us to dump traces whenever outlier requests occur. In the future, we can start dumping more of them. This is an experimental submission to observe its usefulness.

Example:

image

This depends on the following PRs in Osmosis:
- https://github.com/osmosis-labs/osmosis/pull/8254
- https://github.com/osmosis-labs/osmosis/pull/8253


cryptomatictrader:
Any idea how many requests might fall into this 500ms trap? Generating dump can be a high cpu and memory overhead operation, if there are many of them in production, it may have a cascading effect and slow down the entire region.

p0mvn:
Closing in favor of https://github.com/osmosis-labs/sqs/pull/253

PR #221, opened 2024-05-14
POC: local mcap (progress towards dynamic min liquidity)

p0mvn:
ref: https://linear.app/osmosis/issue/STABI-150/dynamic-min-liquidity-filter-to-solve-flashing-routes-problem

Returns local mcap denominated in the denom.

Next, need to add pre-computation logic for pricing it in USDC.

The final goal is to compute the min liquidity value dynamically based on the token input by utilizing these local mcap values.

The query is only exposed for convenience.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -28,6 +28,20 @@
             ],
         },
         {
+            "name": "ingest/usecase",
+            "type": "go",
+            "request": "launch",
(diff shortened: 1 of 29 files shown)


p0mvn:
Closing in favor of https://github.com/osmosis-labs/sqs/pull/224

PR #224, opened 2024-05-15
dynamic min liquidity filter based on token liquidity

p0mvn:
ref: https://linear.app/osmosis/issue/STABI-150/dynamic-min-liquidity-filter-to-solve-flashing-routes-problem

Dynamically chooses the min liquidity capitalization filter based on the minimum of total liquidity capitalization of token in and out.

This is necessary to prevent flashing/bad routes for high-value tokens while allowing the flexible for low-cap tokens to be routable.

See detailed documentation here: https://github.com/osmosis-labs/sqs/pull/281

### Data Ingest Pre-computation

A separate pool liquidity computation worker is created. This worker listens for price updates. Once prices are computed, pool liquidity computation commences.

The final result is that we compute a store mapping from the chain denom to its pool liquidity across all pools as well as pool liquidity capitalization in the default quote denom (USDC)

### Dynamic Min Liquidity Choice

Via config.json, we set a mapping between min token pool liquidity and proposed dynamic min liquidity filter value.

So that if `min(liquidity(token_in_denom), liquidity(token_out_denom)) > config token pool liquidity, we use its associated filter value. If there is no mapping entry present, we fallback to the currently present filter value of `1000`.

At launch, we will have one mapping present.

This incremental rollout will help to ensure the system's stability and empirically validate its success on the highest liquidity pools.

In future, iterations, we can extend the configuration based on production performance.

## Re-architecture

- [x] Pricing updates across blocks are no longer queued up. Instead, their computation is fired off immediately as pricing updates arrive.

## Implementation Plan

- [x] Update pricing worker to not use queue: https://github.com/osmosis-labs/sqs/pull/257
- [x] Store pool denom liquidity capitalization without precision scaling
  * Note: might not be the best idea since we will need descaled values for client APIs. Might instead scale input amounts using usdc price as well
- [x] Store prices for each denom
  * Note: prices are needed if we store pool liquidity values as descaled 
- [x]  Create a function to get dynamic min liquidity filter by converting token in amount into USDC value
   * [x] Higher level APIs & query params
   * [x] Function in tokens usecase 
- [x] Implement observable pattern for pool liquidity pricing worker
- [x] Clean up liquidity min cap filter abstraction: https://linear.app/osmosis/issue/DATA-182/[dynamic-min-liquidity-filter]-refactor-liquidity-pricing-abstraction
   * https://github.com/osmosis-labs/osmosis/pull/8279
   * https://github.com/osmosis-labs/sqs/pull/241

### Unit Testing
- [x] `config.validateDynamicMinLiquidityCapDesc`
- [x] `tokensusecase.GetMinPoolLiquidityCap`
- [x] Pool Liquidity Worker
   * [x] ComputeLiquidityCapitalization
   * [x] hasLaterUpdateThanHeight
   * [x] repriceDenomMetadata
   * [x] OnPricingUpdate 
- [x] New Ingester Methods
   * [x] `updateCurrentBlockLiquidityMapFromBalances`
   * [x] `TransferDenomLiquidityMap`
- [x] Create a task to test higher-level ingester methods using mocks. For now, node testing is OK
- [x] Unit test `ComputeDynamicMinLiquidityCapFilter`
- [x] Unit test `getMinLiquidityCapFilter`
- [x] Unit test `ConvertMinTokensPoolLiquidityCapToFilter`

### Node Testing
- [x] Healthcheck component on a live node
 * Pricing updates propagate
 * Pool liquidity precomputation updates propagate
- [x] Measure if any impact to ingest time

## Observability
- [x] Duration of pool liquidity precomputation

## New Configs

`dynamic-min-liquidity-cap-filters-desc` Implemented as an array. The values represent the min liquidity cap filter values in descending order. This is so that, we can quickly iterate from highest to lowest to find the right value.

The starting value is:
```
"dynamic-min-liquidity-cap-filters-desc": [
        {
            "min-tokens-capitalization": 1000000,
            "filter-value": 75000
        }
]
```
- Means that for any tokenA and tokenB, if the minimum token capitalization of the said tokens is above `1_000_000`, then the `min-liquidity-cap` filter used will be `75000.

Otherwise, we would fallback to the standard `min-pool-liquidity-cap` config value


ingest/usecase/ingest_usecase.go
```diff
@@ -88,8 +90,12 @@ func (p *ingestUseCase) ProcessBlockData(ctx context.Context, height uint64, tak
 
 	// Sort and store pools.
 	p.logger.Info("sorting pools", zap.Uint64("height", height), zap.Duration("duration_since_start", time.Since(startProcessingTime)))
+
 	p.sortAndStorePools(allPools)
 
+	// Update pool denom metadata
+	p.logger.Info("updating pool denom metadata", zap.Uint64("height", height), zap.Int("denom_count", len(uniqueBlockPoolMetadata.DenomPoolLiquidityMap)), zap.Duration("duration_since_start", time.Since(startProcessingTime)))
+
```
p0mvn:
```suggestion
```

router/delivery/http/router_handler.go
```diff
@@ -136,6 +147,37 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
 	return c.JSON(http.StatusOK, quote)
 }
 
+// getMinPoolLiquidityCapFilter returns the min liquidity cap filter for the given tokenIn and tokenOutDenom.
+// if forceDefaultMinLiquidityCap is true, it returns the universal default min pool liquidity capitaliation
+// , ignoring disableMinLiquidityCapFallback.
```
p0mvn:
```suggestion
// if forceDefaultMinLiquidityCap is true, it returns the universal default min pool liquidity capitalization,
// ignoring disableMinLiquidityCapFallback.
```

domain/pricing.go
```diff
@@ -51,6 +51,8 @@ type PricingOptions struct {
 	RecomputePricesIsSpotPriceComputeMethod bool
 	// MinPoolLiquidityCap defines the minimum liquidity required to consider a pool for pricing.
 	MinPoolLiquidityCap uint64
+	// IsWorkerPrecompute defines whether the pricing is precomputed by the worker.
+	IsWorkerPrecompute bool
```
cryptomatictrader: Better naming, like ```IsWorkerPrecomputePricing```, or ```IsPricingPrecomputedByWorker```?

cryptomatictrader (approved):
Looks great! Might just want to rename the variable a bit for better readability. Thanks!

chaininfo/usecase/chain_info_usecase.go
```diff
@@ -20,28 +20,35 @@ type chainInfoUseCase struct {
...
 const (
 	MaxAllowedHeightUpdateTimeDeltaSecs = 30
 
-	// Number of heights of buffer between the latest state height and the latest price update height
-	// We fail the healtcheck if the difference between the two becomes greater than this constant.
-	priceUpdateHeightBuffer  = 50
-	initialPriceUpdateHeight = 0
+	// Number of heights of buffer between the latest state height and the latest price/pool liquidity update height
+	// We fail the healtcheck if the difference between the current and last becomes greater than this constant.
+	updateHeightBuffer  = 50
```
cryptomatictrader: Based on the comment, ```updateHeightBuffer``` seems to be used for threshold and alert purposes, shall we rename it something like ```updateHeightThreshold```? Or this is also used for some buffering purposes.

chaininfo/usecase/chain_info_usecase.go
```diff
@@ -97,20 +104,46 @@ func (p *chainInfoUseCase) OnPricingUpdate(ctx context.Context, height uint64, b
...
+	// Check that the initial pool liquidities have been computed and received.
+	if currentUpdateHeight == initialUpdateHeight {
+		return fmt.Errorf("healthcheck has not received initial %s updates", updateName)
 	}
 
-	// Check that the price updates have been occurring
-	if latestPriceUpdateHeight < p.lastSeenUpdatedHeight-priceUpdateHeightBuffer {
-		return errors.New("latest price update height is less than the last seen updated height")
+	// Check that the pool liquidity updates have been occurring
+	if currentUpdateHeight < latestIngestedHeight-updateHeightBuffer {
```
cryptomatictrader: Space before and after the minus ```-``` sign for readability?

cryptomatictrader (approved):
Thank you! Since we are adding endpoints, shall we also update sqs README to include those endpoints?

chaininfo/usecase/chain_info_usecase.go
p0mvn: This is a standard go formatting. It would auto-set it back to the current way if I were to change this.

Let's generally aim to set up linters for any style-related rules

chaininfo/usecase/chain_info_usecase.go
p0mvn: To me, the semantics feel equivalent but I applied the recommendation here: [5e42c1a](https://github.com/osmosis-labs/sqs/pull/224/commits/5e42c1a432d1614e77d444b352a4d70ab1f505e1)

domain/pricing.go
p0mvn: Done [c399034](https://github.com/osmosis-labs/sqs/pull/224/commits/c39903440ec782d2389fbc51724c3586f97147c6)

p0mvn:
> Thank you! Since we are adding endpoints, shall we also update sqs README to include those endpoints?

Thanks for the review.

Let's use swagger as documentation for the endpoints. There is no reason to fragment the documentation to README.

The README docs predate swagger but I will remove it at some point to avoid confusion from outdated context

PR #225, opened 2024-05-16
DATA-68 [sqs-automated-tests-in-ci][prices]

cryptomatictrader:
@p0mvn , I found test results for low vol or low liq are not very deterministic. In most of the time, they fail as they don't have coingecko id, or coingecko price even with the coingecko id, and very high price discrepancy (>0.8) between sqs and coingecko pricing. We should consider skipping low vol or low liq test cases if they are run in the critical deployment or alert path.


p0mvn:
I agree, and I think that it is sufficient to only run high liquidity tests against coingecko prices.

Contrary to chain tests, our goal is only to validate that Coingecko prices work in general. For the support of tokens, we are depending on CoinGecko so the granularity in terms of liq/volume is not very important as it is not under our control.

Also, the error handling of Coingecko IDs not being present should be handled by unit tests at a lower level of abstraction

tests/sqs_service.py
```diff
@@ -62,4 +67,40 @@ def get_tokens_metadata(self):
...
-        return self.tokens_metadata
\ No newline at end of file
+        return self.tokens_metadata
+
+    # Given the base and quote denom, call the SQS API endpoint /tokens/prices
+    # and return the token price in the response
+    def get_tokens_prices(self, denom, quote_denom, human_denoms="false"):
+        # Set the query parameters
+        params = {
+            "base": denom,
```
p0mvn: We have support for multiple base denoms (and will have support for multiple quotes in the future). Could we expand the API to support multiple tokens?

And also add test cases with multiple denoms submitted at once

tests/sqs_service.py
```diff
@@ -62,4 +67,40 @@ def get_tokens_metadata(self):
...
+                raise Exception(f"Error fetching asset list: {response.text}")
+            asset_list_json = response.json()
+            for asset in asset_list_json.get("assets"):
+                self.asset_list[asset[coin_minimal_denom_key]] = asset.get(coingecko_id_key, None)
+            
+        return self.asset_list.get(denom, None) 
+
+    
+
+
```
p0mvn:
```suggestion
```

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,65 @@
...
+    @pytest.mark.parametrize("token",setup.choose_tokens_volume_range(NUM_TOKENS_DEFAULT, MIN_VOL_FILTER_DEFAULT, MAX_VAL_MID_VOL_FILTER_DEFAULT))
+    def test_mid_volume_token_prices(self, environment_url, token):
+        self.run_coingecko_comparison_test(environment_url, token, MID_PRICE_DIFF)
+
+    # NUM_TOKENS_DEFAULT top by-volume tokens
+    @pytest.mark.parametrize("token", setup.choose_tokens_volume_range(NUM_TOKENS_DEFAULT))
+    def test_top_volume_token_prices(self, environment_url, token):
+        self.run_coingecko_comparison_test(environment_url, token, LOW_PRICE_DIFF)
+
+    def run_coingecko_comparison_test(self, environment_url, token, price_diff_threshold):
```
p0mvn: This is great!

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,65 @@
...
+import time
+
+from datetime import datetime
+from sqs_service import *
+from coingecko_service import *
+from conftest import SERVICE_MAP
+from conftest import SERVICE_COINGECKO
+from constants import *
+
+class TestTokensPrices:
```
p0mvn, line 18: I would like to suggest the following test cases:
- All valid listed tokens as defined by [this function](https://github.com/osmosis-labs/sqs/blob/3c66c5f7fe24c223fd73ff3cd9c0b1adefc69ff6/tests/setup.py#L263)
  * Track all the prices that we do not support and print them
  * Assert that the number of unsupported tokens is below some pre-configured threshold. I think that 30 is a good starting value
- Multiple base denoms requested in a single query.

Additionally, let's also add support for the SQS CoinGecko integration price test (can be a separate PR) - just noting so that we don't forget

tests/test_tokens_prices.py
p0mvn: Could we add a summary of what this helper does?

Similar to:
https://github.com/osmosis-labs/sqs/blob/3c66c5f7fe24c223fd73ff3cd9c0b1adefc69ff6/tests/setup.py#L259-L260

p0mvn (commented):
Nice job!

tests/sqs_service.py
cryptomatictrader: Good idea. Done.

tests/test_tokens_prices.py
cryptomatictrader: Test cases added as suggested and separate PR created as it depends on coingecko integration PR merge.

tests/test_tokens_prices.py
cryptomatictrader: Done

cryptomatictrader:
@p0mvn , feel free to comment and let me know when it can be merged into v25.x. Thanks!

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,124 @@
...
+        sqs_price_json = sqs_service.get_tokens_prices(tokens)
+        for token in tokens:
+            sqs_price_str = sqs_price_json.get(token, {}).get(USDC, None)
+            assert sqs_price_str is not None, f"{token} SQS price is none"
+            sqs_price = float(sqs_price_str)
+            assert sqs_price > 0, f"{token} SQS price is zero"
+
+    # Test every valid listed token if it is supported by the /tokens/prices endpoint
+    # Count the number of unsupported tokens and assert it is below the threshold
+    def test_unsupported_token_count(self, environment_url):
```
p0mvn: Could we parameterize this test case similar to others so that we can run them in parallel while at the same time having the assertion about the unsupported count?

We could have an atomic integer in the global space that each test case would update and then a separate test fixture that would assert on its value relative to the threshold.

As of right now, we would be running each in sequence, increasing the CI time

p0mvn (commented):
Nice job!

Had one suggestion.

Additionally, I noticed some tests failing on my side:
```
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[https://sqs.stage.osmosis.zone-ibc/CFF40564FDA3E958D9904B8B479124987901168494655D9CC6B7C0EC0416020B] - AssertionError: ibc/CFF40564FDA3E958D9904B8B479124987901168494655D9CC6B7C0EC0416020B coingecko id is none
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_volume_token_prices[https://sqs.stage.osmosis.zone-factory/osmo19hdqma2mj0vnmgcxag6ytswjnr8a3y07q7e70p/wLIBRA] - AssertionError: factory/osmo19hdqma2mj0vnmgcxag6ytswjnr8a3y07q7e70p/wLIBRA coingecko id is none
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[https://sqs.stage.osmosis.zone-ibc/7FA7EC64490E3BDE5A1A28CBE73CC0AD22522794957BC891C46321E3A6074DB9] - AssertionError: ibc/7FA7EC64490E3BDE5A1A28CBE73CC0AD22522794957BC891C46321E3A6074DB9 coingecko id is none
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[https://sqs.stage.osmosis.zone-ibc/346786EA82F41FE55FAD14BF69AD8BA9B36985406E43F3CB23E6C45A285A9593] - AssertionError: ibc/346786EA82F41FE55FAD14BF69AD8BA9B36985406E43F3CB23E6C45A285A9593,darcmatter-coin price difference (0.48469608344216497) is gre...
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[https://sqs.stage.osmosis.zone-ibc/F16FDC11A7662B86BC0B9CE61871CBACF7C20606F95E86260FD38915184B75B4] - AssertionError: ibc/F16FDC11A7662B86BC0B9CE61871CBACF7C20606F95E86260FD38915184B75B4 coingecko id is none
```

Let's try to eliminate flakiness prior to merge :pray:

tests/test_tokens_prices.py
cryptomatictrader: It looks like pytest-xdist runs the tests in different processes but not threads and so we need to do some inter-process communication. I can think of using FileLock to do so. Do you have some better idea?

p0mvn: Got it, interesting.

From my investigations, most suggestions propose similar solutions.

We could then simply spawn Python threads within a single test case. While not parallel, we would still benefit from concurrency of not having to wait for blocking network requests, improving CI times.

Let me know what you think

cryptomatictrader: Issues resolved. Please check out the latest commit. Thanks!

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,156 @@
...
+    # Tests are run by separate processes in parallel, thus using the filelock
+    # to ensure that the counter is updated safely 
+    @pytest.mark.parametrize("token", setup.choose_valid_listed_tokens())
+    def test_unsupported_token_count(self, environment_url, token):
+        sqs_service = SERVICE_MAP[environment_url]
+        try:
+            sqs_price_json = sqs_service.get_tokens_prices([token])
+        except Exception as e:
+            # Increment unsupported token count if an exception is raised
+            self.increment_counter
```
p0mvn:
```suggestion
            self.increment_counter()
```

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,156 @@
...
+            f"Unsupported token {token}: error fetching sqs price {str(e)}"
+        sqs_price_str = sqs_price_json.get(token, {}).get(USDC, None)
+        if sqs_price_str is None:
+            self.increment_counter
```
p0mvn:
```suggestion
            self.increment_counter()
```

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,156 @@
...
+            # Increment unsupported token count if the price is not available
+            f"Unsupported token {token}: SQS price is none in response"
+        sqs_price = float(sqs_price_str)
+        if sqs_price <= 0:
+            self.increment_counter
```
p0mvn:
```suggestion
            self.increment_counter()
```

tests/test_tokens_prices.py
```diff
@@ -0,0 +1,156 @@
...
+            os.remove(counter_file)
+        self.write_counter(0)
+
+    # Assert that the unsupported token count is within the threshold
+    # Clean up the counter file at the end of the test
+    def teardown_method(self):
+        unsupported_token_count = self.read_counter()
+        assert unsupported_token_count <= UNSUPPORTED_TOKEN_COUNT_THRESHOLD, f"Unsupported token count: {unsupported_token_count} exceeds threshold {UNSUPPORTED_TOKEN_COUNT_THRESHOLD}"
+        if os.path.exists(counter_file):
+            os.remove(counter_file)
```
p0mvn: I belive that `setup_method` runs before every price test and `teardown_method` after every price test.

As an outcome, the `setup_method` resets the counter to zero.

p0mvn (changes requested):
Tried running the unsupported suite with debugger and noticed some issues.

Leaving some comments about what I think might be the problem. Please take a look.

Overall, I like the FileLock approach. Thanks for setting it up!

cryptomatictrader:
Thanks a lot for catching this! Revised it and ran positive and negative cases and it should be good now!

p0mvn (approved):
🚀

PR #231, opened 2024-05-17
WIP: orderbook pool integration

crnbarr93:
These changes are to integrate orderbook CosmWasm pools in to SQS. 

The orderbook smart contract already implements the CW Pool interface, however to help with efficiency the calculation for `CalculateTokenOutByTokenIn` is done by SQS itself. This calculation requires awareness of all tick states from each orderbook contract.


router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,188 @@
...
+			tickIdx--
+		} else {
+			return sdk.Coin{}, sqsdomain.OrderbookPoolInvalidDirectionError{Direction: direction}
+		}
+
+		// Calculate the price for the current tick
+		tickPrice, err := clmath.TickToPrice(tick.TickId)
+		if err != nil {
+			return sdk.Coin{}, err
+		}
```
p0mvn, line 102: Is our order book tick-to-price conversion the same as CL?

sqsdomain/orderbook.go
```diff
@@ -0,0 +1,60 @@
+package sqsdomain
```
p0mvn: `sqsdomain` is a separate module that is shared between `sqs` and `osmosis`.

If we do not intend to have the `osmosis` node utilize these APIs, I suggest moving this file into `domain` package instead that is `sqs` specific. This would help reduce dependency management overhead

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+
+func amountToValue(amount osmomath.BigDec, price osmomath.BigDec, direction int64) osmomath.BigDec {
+	if direction == sqsdomain.ASK {
+		return amount.Mul(price)
+	} else {
+		return amount.Quo(price)
+	}
+}
```
p0mvn: Can `Mul` or `Quo` be mutative?

router/usecase/pools/routable_cw_orderbook_pool.go
AlpinYukseloglu: Yes, but reimplemented in rust so there might be minor differences due to precision diffs

sqsdomain/orderbook.go
crnbarr93: Moved references to `domain` instead of `sqsdomain`

router/usecase/pools/routable_cw_orderbook_pool.go
crnbarr93: Altered both to be mutative

p0mvn:
Closing as this have been merged in a separate PR

PR #241, opened 2024-05-18
refactor: clean min liquidity capitalization filter abstraction

p0mvn:
Progress towards: https://linear.app/osmosis/issue/DATA-182/[dynamic-min-liquidity-filter]-refactor-liquidity-pricing-abstraction

## Context

Due to legacy & initial dev velocity reasons, we have the liquidity pricing abstraction identified as OSMO in some places in USDC in others.

It is priced in OSMO but we want it to be in USDC.

However, we should instead make the abstraction more general and have it configured via a config value.

Let's consistently change the name to:

TotalLiquidityCap where "Cap" signified capitalization

We currently do pricing in the osmosis node but that is to eventually be migrated to sqs. As a result for now, we can just hardcode a pricing pool between OSMO and USDC.

In the future, we will simply configure a quote in sqs for pricing and use the router to figure out the best path

## Notes

- Will require updates on the chain side and deploying the changes at once:
   * Ref: https://github.com/osmosis-labs/osmosis/pull/8279
- Will require updating infra configs
   * Ref: https://github.com/osmosis-labs/infrastructure/pull/247


README.md
```diff
@@ -574,9 +574,8 @@ See the recommended enabled configuration below:
       "max-split-routes": 3,
       // Maximum number of iterations to split a route across.
       "max-split-iterations": 10,
-      // Minimum liquidity for a pool to be included in the router
-      // denominated in OSMO.
-      "min-osmo-liquidity": 100,
+      // Minimum liquidity cap for a pool to be considered in the router.
```
p0mvn:
```suggestion
      // Minimum liquidity capitalization for a pool to be considered in the router.
```

domain/pricing.go
```diff
@@ -103,8 +105,8 @@ type PricingConfig struct {
 
 	MaxPoolsPerRoute int `mapstructure:"max-pools-per-route"`
 	MaxRoutes        int `mapstructure:"max-routes"`
-	// Denominated in OSMO (not uosmo)
-	MinOSMOLiquidity int `mapstructure:"min-osmo-liquidity"`
+	// MinPoolLiquidityCapFilter is the minimum liquidity required for a pool to be considered in the router.
```
p0mvn:
```suggestion
	// MinPoolLiquidityCap is the minimum liquidity capitalization required for a pool to be considered in the router.
```

domain/router.go
```diff
@@ -70,8 +70,8 @@ type RouterConfig struct {
 	MaxRoutes          int      `mapstructure:"max-routes"`
 	MaxSplitRoutes     int      `mapstructure:"max-split-routes"`
 	MaxSplitIterations int      `mapstructure:"max-split-iterations"`
-	// Denominated in OSMO (not uosmo)
-	MinOSMOLiquidity          int  `mapstructure:"min-osmo-liquidity"`
+	// MinPoolLiquidityCap is the minimum liquidity required for a pool to be considered in the route.
```
p0mvn:
```suggestion
	// MinPoolLiquidityCap is the minimum liquidity capitalization required for a pool to be considered in the route.
```

README.md
```diff
@@ -574,9 +574,8 @@ See the recommended enabled configuration below:
       "max-split-routes": 3,
       // Maximum number of iterations to split a route across.
       "max-split-iterations": 10,
-      // Minimum liquidity for a pool to be included in the router
-      // denominated in OSMO.
-      "min-osmo-liquidity": 100,
+      // Minimum liquidity capitalization for a pool to be considered in the router.
```
cryptomatictrader, line 592:
```suggestion
      // Minimum liquidity capitalization in USDC for a pool to be considered in the router.
```

app/sqs_config.go
```diff
@@ -25,7 +25,7 @@ var DefaultConfig = domain.Config{
 		MaxRoutes:                        5,
 		MaxSplitRoutes:                   3,
 		MaxSplitIterations:               10,
-		MinOSMOLiquidity:                 100, // 100 OSMO
+		MinPoolLiquidityCap:              100,
```
cryptomatictrader:
```suggestion
		MinPoolLiquidityCap:              100, // 100 USDC
```

README.md
p0mvn: I would like to keep this more general because it is to be eventually set by the following config:
https://github.com/osmosis-labs/sqs/blob/187ad1b75d6605b7ec80bec4601e77b49c8d7648/config.json#L37

So it might not always be USDC, at some point we might switch to USDT.

I've aimed to explain this in a few places but let me also add some comments here:
[0f3ccd3](https://github.com/osmosis-labs/sqs/pull/241/commits/0f3ccd3fcb7c69949c599bc0fc4df2b0fc3b3e4e)

cryptomatictrader: Small suggestion - I'd like to know the unit when making changes to a config

app/sqs_config.go
cryptomatictrader: Same

p0mvn: Ditto:
https://github.com/osmosis-labs/sqs/pull/241#discussion_r1607475802

[79d9f49](https://github.com/osmosis-labs/sqs/pull/241/commits/79d9f4972038d0d187edb61b24f3519ee84b5019)

p0mvn:
@cryptomatictrader thanks for the comments.

Addresses & resolved conflicts

cryptomatictrader (approved):
Looks good to me. The general suggestion I'd like to make is to add some comments in areas to hint the unit is USDC, as sometimes it makes me nervous when making config changes without 100% sure what unit it is referring to. Thanks!

README.md
p0mvn: Does the comment [I added](https://github.com/osmosis-labs/sqs/blob/f6a175fa550d2aa765b3be01280ed0639fd54a27/README.md?plain=1#L578) about the [pricing.default-quote-human-denom](https://github.com/osmosis-labs/sqs/blob/f6a175fa550d2aa765b3be01280ed0639fd54a27/README.md?plain=1#L604) address your request? 


From that comment, it should be clear it is USDC.

However, I'll apply the recommendation if you think the change I made is insufficient

p0mvn:
> Looks good to me. The general suggestion I'd like to make is to add some comments in areas to hint the unit is USDC, as sometimes it makes me nervous when making config changes without 100% sure what unit it is referring to. Thanks!

Thanks. Yep - good suggestion.

The default quote denom to be used universally across the system is set by this config:
https://github.com/osmosis-labs/sqs/blob/f6a175fa550d2aa765b3be01280ed0639fd54a27/config.json#L37

Applied additional comments in the suggested areas

p0mvn:
We will merge this after: https://github.com/osmosis-labs/sqs/pull/209

PR #245, opened 2024-05-22
Alloy Transmuter Support

iboss-ptk:
- Add support for alloy transmuter
- Use cw2 info to determine the pool type
- Pass along `CosmWasmPoolModel` with information queried from contract to sqs
- Fix CosmWasm pool conversion failed on streaming update due to missing wasmkeeper

sqs tag for these changes block https://github.com/osmosis-labs/osmosis/pull/8291


sqsdomain/pools.go
```diff
@@ -36,6 +36,15 @@ type PoolI interface {
 	// If this is not a concentrated pool, errors
 	SetTickModel(*TickModel) error
 
+	// GetCWPoolModel returns the CosmWasm pool model for the pool
+	// if this is a CosmWasm pool. Errors otherwise
+	// No error if this is a CosmWasm pool but the model is not set
+	GetCWPoolModel() (*CWPoolModel, error)
```
p0mvn: nit: to make it clearer for any future new contributors
```suggestion
	GetCosmWasmPoolModel() (*CWPoolModel, error)
```

sqs.log
```diff
@@ -1,665 +1,1472 @@
-{"level":"info","ts":1713407057.397006,"msg":"log level","is_debug":false,"log_level":"info"}
```
p0mvn: nit: do you mind deleting this file before final review? 🙏

sqsdomain/pools.go
```diff
@@ -60,9 +69,10 @@ type SQSPool struct {
 }
 
 type PoolWrapper struct {
-	ChainModel poolmanagertypes.PoolI `json:"underlying_pool"`
-	SQSModel   SQSPool                `json:"sqs_model"`
-	TickModel  *TickModel             `json:"tick_model,omitempty"`
+	ChainModel  poolmanagertypes.PoolI `json:"underlying_pool"`
+	SQSModel    SQSPool                `json:"sqs_model"`
+	TickModel   *TickModel             `json:"tick_model,omitempty"`
+	CWPoolModel *CWPoolModel           `json:"cw_pool_model,omitempty"`
```
p0mvn: Can we instead define `*CWPoolModel` on `SQSModel`?

That way, we can avoid defining concrete CosmWasm getter and setter APIs on the generalized `PoolI` interface, helping to keep it lean.

Pet this suggestion, we would be able to use `GetSQSPoolModel().CWPoolModel`

p0mvn (commented):
Just did a quick skim without any deep review for now. Looks like everything is going well and on the right track. Great work!

sqs.log
iboss-ptk: noted

sqsdomain/pools.go
iboss-ptk: sounds good! will do that

sqsdomain/cw_pool.go
```diff
@@ -0,0 +1,52 @@
...
+		ContractInfo: ContractInfo{
+			Contract: contract,
+			Version:  version,
+		},
+		Data: data,
+	}
+}
+
+func (model *CosmWasmPoolModel) IsAlloyTransmuter() bool {
+	return model.ContractInfo.Matches("crates.io:transmuter", "3.0.0")
```
p0mvn: Would it make sense to set this via config?

sqsdomain/cw_pool.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type CosmWasmPoolModel struct {
+	ContractInfo ContractInfo `json:"contract_info"`
+	Data         CWPoolData   `json:"data"`
+}
+
+type CWPoolData struct {
+	AlloyTransmuter *AlloyTransmuterData `json:"alloy_transmuter,omitempty"`
+}
```
p0mvn: nit: could we add godocs to all types. I'm sure we would benefit from these in the future

router/usecase/router.go
```diff
@@ -65,10 +65,14 @@ func ValidateAndSortPools(pools []sqsdomain.PoolI, cosmWasmPoolsConfig domain.Co
 				continue
 			}
 
+			logger.Debug("cw pool code id is not added to config, skip silently", zap.Uint64("pool_id", pool.GetId()))
```
p0mvn: This log does not make sense to me to be in this location

router/usecase/pools/pool_factory.go
```diff
@@ -96,20 +96,47 @@ func newRoutableCosmWasmPool(pool sqsdomain.PoolI, cosmWasmConfig domain.CosmWas
...
+			return &routableAlloyTransmuterPoolImpl{
+				ChainPool:           cosmwasmPool,
+				AlloyTransmuterData: model.Data.AlloyTransmuter,
+				Balances:            balances,
+				TokenOutDenom:       tokenOutDenom,
+				TakerFee:            takerFee,
+				SpreadFactor:        spreadFactor,
+			}, nil
+		}
+	}
```
p0mvn: I think we should still have the ability to control whether we support an alloyed transmitter by having a config for them somewhere [here](https://github.com/osmosis-labs/sqs/blob/7f9fdda008ac7dd6ab599caec64622d375543d44/config.json#L20).

In previous cw pool releases, we had several unexpected issues arise so controlling what is supported by our router proved to be useful.

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,166 @@
...
+func (*routableAlloyTransmuterPoolImpl) IsGeneralizedCosmWasmPool() bool {
+	return false
+}
+
+// GetCodeID implements sqsdomain.RoutablePool.
+func (r *routableAlloyTransmuterPoolImpl) GetCodeID() uint64 {
+	return r.ChainPool.CodeId
+}
+
+func (r *routableAlloyTransmuterPoolImpl) findNormalizationFactors(tokenInDenom, tokenOutDenom string) (osmomath.Int, osmomath.Int, error) {
```
p0mvn: Could you link some docs to https://github.com/osmosis-labs/transmuter that explain normalization factors please?

I think that future contributors would benefit from this

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return osmomath.ZeroBigDec(), err
+	}
+
+	tokenInAmount := osmomath.NewBigDec(tokenIn.Amount.Int64())
+
+	tokenInNormFactorBig := osmomath.NewBigIntFromBigInt(tokenInNormFactor.BigInt())
+	tokenOutNormFactorBig := osmomath.NewBigIntFromBigInt(tokenOutNormFactor.BigInt())
+
+	return tokenInAmount.MulInt(tokenOutNormFactorBig).QuoInt(tokenInNormFactorBig), nil
```
p0mvn, line 176: Can we handle unexpected truncation to zero here?

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/osmosis-labs/sqs/domain"
+	"github.com/osmosis-labs/sqs/domain/mocks"
+	"github.com/osmosis-labs/sqs/router/usecase/pools"
+	"github.com/osmosis-labs/sqs/sqsdomain"
+
+	"github.com/osmosis-labs/osmosis/osmomath"
+)
+
+// Tests no slippage quotes and validation edge cases aroun transmuter pools.
+func (s *RoutablePoolTestSuite) TestCalculateTokenOutByTokenIn_AlloyTransmuter() {
```
p0mvn, line 62: Could we add a test for truncation in division occurring here?

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
p0mvn: Would be great to unit test each sub-method on the calc path individually - the happy paths and major error/edge cases 🙏

p0mvn (commented):
Great work! Left some comments

p0mvn:
Also, would be great to document somewhere the logic around how `CWPoolModel` relies on the contract info and how this is different from transmuter and generalized cw pools. Maybe we could start a doc under `docs/architecture/CW_POOLS.MD`?

sqsdomain/cw_pool.go
iboss-ptk: I don't think so as this function is referenced by ingester codebase as well. Also this information will not be varied from environment to environment.

p0mvn:
@iboss-ptk, please ping me once ready for another round of 👀

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
iboss-ptk: Not sure which part would you expect truncation to happen? Are you referring to high `tokenInNormFactor` that might turn result into 0?  What would an expected behavior in such case would you like it to be?

domain/router.go
```diff
@@ -100,6 +100,9 @@ type PoolsConfig struct {
 	// NOTE: that these pools make network requests to chain for quote estimation.
 	// As a result, they are excluded from split routes.
 	GeneralCosmWasmCodeIDs []uint64 `mapstructure:"general-cosmwasm-code-ids"`
+  
+  // Feature flag for alloyed transmuter
+  IsAlloyedTransmuterEnabled bool `mapstructure:"is-alloyed-transmuter-enabled"`
```
p0mvn: side nit: could you format this to be on the same indent with other fields?

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
p0mvn: Thinking more about this - I think it is fine if we truncate. In that case, the value is probably small to the extent of being insignificant

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,169 @@
...
```
p0mvn, line 176: Does it make sense to add a division by zero check?

p0mvn: Could we also maximize the usage of mutative (suffixed with `Mut`) versions of the math operations for performance reasons?

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
p0mvn: I think the test case for this is still applicable.

We can choose the denominator large enough and numerator small enough to take the final result to zero despite the starting inputs being non-zero

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	for name, tc := range tests {
+		s.Run(name, func() {
+			s.Setup()
+
+			cosmwasmPool := s.PrepareCustomTransmuterPool(s.TestAccs[0], []string{USDC, USDT})
+
+			poolType := cosmwasmPool.GetType()
+
+			veryBigNormalizationFactor, _ := osmomath.NewIntFromString("999999999999999999999999999999999999999999")
```
p0mvn: nit: can we define this at the top of the test fixture to avoid reinitializing for every single one

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,275 @@
...
+						},
+					},
+				),
+				PoolType: poolType,
+			}
+
+			routablePool, err := pools.NewRoutablePool(mock, USDT, noTakerFee, domain.CosmWasmPoolRouterConfig{
+				IsAlloyedTransmuterEnabled: true,
+			})
+			s.Require().NoError(err)
```
p0mvn: I understand why we're choosing to use this boilerplate but could we either add comments or encapsulate this in a separate function for readability?

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,275 @@
...
+				s.Require().NoError(err)
+				s.Require().Equal(tc.expectedInNormFactor, inNormFactor)
+				s.Require().Equal(tc.expectedOutNormFactor, outNormFactor)
+			}
+		})
+	}
+}
+
+func (s *RoutablePoolTestSuite) TestCalcTokenOutAmt_AlloyTransmuter() {
+	tests := map[string]struct {
```
p0mvn, line 200: Can we add a test case where the normalization factor is zero to prompt a division by zero and panic in the final calculation?

We should probably handle that gracefully and return an error instead

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
p0mvn: We could also reuse the initialisation helped across different test cases

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,169 @@
...
+
+// GetCodeID implements sqsdomain.RoutablePool.
+func (r *routableAlloyTransmuterPoolImpl) GetCodeID() uint64 {
+	return r.ChainPool.CodeId
+}
+
+// FindNormalizationFactors finds the normalization factors for the given token in and token out denoms.
+// It is required for calculating token out & spot price.
+// For more information about normalization factor, please refer to [transmuter documentation](https://github.com/osmosis-labs/transmuter/tree/v3.0.0?tab=readme-ov-file#normalization-factors).
+func (r *routableAlloyTransmuterPoolImpl) FindNormalizationFactors(tokenInDenom, tokenOutDenom string) (osmomath.Int, osmomath.Int, error) {
```
p0mvn, line 124: I like how you added a separate test for a lower-level calc method but could we also unit-test this one?

Seems like a few things are happening here, and we would definitely benefit from strict coverage

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
iboss-ptk: noted

iboss-ptk: `normalization_factor` should never be 0 and it's validated in contract. What would fail gracefully looks like for you? explicitly return error?

iboss-ptk: It seems that `Mut` variants does not have `MulInt` and `QuoInt` but only `MulMut` and `QuoMut` which takes Dec and will loose precision if converting those number to Dec. Do we want that for performance?

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
iboss-ptk: There are already unit-test cases for this function.

config.json
```diff
@@ -32,7 +32,8 @@
             572,
             773,
             641
-        ]
+        ],
+        "is-alloyed-transmuter-enabled": true
```
p0mvn: Can you please update the following and merge them before this PR is merged?

Template: https://github.com/osmosis-labs/infrastructure/blob/55a5946330250d38047d26f8c4865474436db9d3/environments/sqs-osmosis-zone/ansible/roles/prepare_local_folder/templates/sqs/config.json.j2

Configs:
- https://github.com/osmosis-labs/infrastructure/blob/main/environments/sqs-osmosis-zone/environments/stage/ansible/vars/sqs.yaml
- https://github.com/osmosis-labs/infrastructure/blob/main/environments/sqs-osmosis-zone/environments/prod/ansible/vars/sqs.yaml
- https://github.com/osmosis-labs/infrastructure/blob/main/environments/sqs-osmosis-zone/environments/testnet/ansible/vars/sqs.yaml

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,253 @@
...
+			tokenOutDenom:    "INVALID",
+			expectedTokenOut: osmomath.ZeroBigDec(),
+			expectedError:    domain.MissingNormalizationFactorError{Denom: "INVALID", PoolId: 1},
+		},
+		"missing normalization factors for both token in and token out": {
+			tokenIn:          sdk.NewCoin("INVALID", osmomath.NewInt(100)),
+			tokenOutDenom:    "INVALID",
+			expectedTokenOut: osmomath.ZeroBigDec(),
+			expectedError:    domain.MissingNormalizationFactorError{Denom: "INVALID", PoolId: 1},
+		},
```
p0mvn: nit: can we define constants for the same denoms. Would make test updates simpler in the future

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,253 @@
...
+		CosmWasmPoolModel: sqsdomain.NewCWPoolModel(
+			"crates.io:transmuter", "3.0.0",
+			sqsdomain.CWPoolData{
+				AlloyTransmuter: &sqsdomain.AlloyTransmuterData{
+					AlloyedDenom: ALLUSD,
+					AssetConfigs: []sqsdomain.TransmuterAssetConfig{
+						{Denom: USDC, NormalizationFactor: osmomath.NewInt(100)},
+						{Denom: USDT, NormalizationFactor: osmomath.NewInt(1)},
+						{Denom: "overlypreciseusd", NormalizationFactor: veryBigNormalizationFactor},
+						{Denom: "noprecisionusd", NormalizationFactor: osmomath.ZeroInt()},
```
p0mvn: re: the previous comment about constants. I think by having them referenced via shared variables, noting such dependent setup logic would be simpler since we could search variables by references

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	tests := map[string]struct {
+		tokenIn     sdk.Coin
+		tokenOut    sdk.Coin
+		balances    sdk.Coins
+		expectError error
+	}{
+		"valid transmuter quote": {
+			tokenIn:  sdk.NewCoin(USDT, osmomath.NewInt(10000)),
+			tokenOut: sdk.NewCoin(USDC, osmomath.NewInt(1000000)),
```
p0mvn: nit: ditto for vars for int values to simplify future updates and reference checking.

Applies to other tests as well.

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,177 @@
...
+		return osmomath.ZeroBigDec(), err
+	}
+
+	if tokenInNormFactor.IsZero() {
+		return osmomath.ZeroBigDec(), domain.ZeroNormalizationFactorError{Denom: tokenIn.Denom, PoolId: r.GetId()}
+	}
+
+	if tokenOutNormFactor.IsZero() {
+		return osmomath.ZeroBigDec(), domain.ZeroNormalizationFactorError{Denom: tokenOutDenom, PoolId: r.GetId()}
+	}
```
p0mvn: In Go, it is idiomatic to return a nil value rather than an initialized zero when an error is returned.

Would it make sense to return an uninitialized `osmomath.BigDec{}` here for consistency

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -0,0 +1,177 @@
...
+// - the token in amount is greater than the balance of the token out
+//
+// Note that balance validation does not apply to alloyed asset since it can be minted or burned by the pool.
+func (r *routableAlloyTransmuterPoolImpl) CalculateTokenOutByTokenIn(ctx context.Context, tokenIn sdk.Coin) (sdk.Coin, error) {
+	tokenOutAmt, err := r.CalcTokenOutAmt(tokenIn, r.TokenOutDenom)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	tokenOutAmtInt := osmomath.NewIntFromBigInt(tokenOutAmt.TruncateInt().BigInt())
```
p0mvn: Could we do `tokenOutAmt.Dec().TruncateInt()` to simplify the conversion?

sqsdomain/cw_pool.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Configuration for each asset in the transmuter pool
+type TransmuterAssetConfig struct {
+	// Denom of the asset
+	Denom string `json:"denom"`
+
+	// Normalization factor for the asset.
+	// [more info](https://github.com/osmosis-labs/transmuter/tree/v3.0.0?tab=readme-ov-file#normalization-factors)
+	NormalizationFactor osmomath.Int `json:"normalization_factor"`
```
p0mvn, line 81: Can these be regular Go ints?

If so, we would save on math operations and serialization overhead while also improving readability from not having to do `osmomath.Int` to `osmomath.BigDec` conversions.

p0mvn (commented):
This already looks good to me. Nice work!

However, I would like to run the tests before approving. Will come back to this.

Btw please make PRs from this repo directly (not fork) - it would simplify the review process in the future.

I will add you to the repo

sqsdomain/cw_pool.go
```diff
@@ -0,0 +1,65 @@
...
+		ContractInfo: ContractInfo{
+			Contract: contract,
+			Version:  version,
+		},
+		Data: data,
+	}
+}
+
+func (model *CosmWasmPoolModel) IsAlloyTransmuter() bool {
+	return model.ContractInfo.Matches("crates.io:transmuter", "3.0.0")
```
p0mvn: What if the contract is migrated? We would then require code updates and redeployment?

Please consider removing this hard-coded dependency here and on the node side.

The current code ID model allows us to simply update the config without performing any code changes

p0mvn: Please consider going back to the code ID config when determining if a pool is an alloy transmuter within SQS. 

To achieve this on the Osmosis node side, we can continue utilizing the CW info but only match the name e.g. `crates.io:transmuter-alloy` (no version). This way, if the version is updated, it will still be auto-propagated from Osmosis node. However, give SQS the flexibility of handling it via standard means.

The proposed code ID approach would ensure compatibility with how other cw pool types are handled while also allowing the flexibility of only updating configs without code changes if migration is done.

With the current approach, we still need a boolean config but also hardcode the version here which I think is suboptimal for operations.

Please let me know what you think

sqsdomain/cw_pool.go
iboss-ptk: It can't, normalization factor query response is represented as string due to potential misprecision in javascript client. So the responded value is not just an int.

sqsdomain/cw_pool.go
iboss-ptk: noted

iboss-ptk: but then, this logic required on the ingester side, do we also have config on that side, I didn't see any. If there is any update required, both side needs to be effected.

iboss-ptk: I think versions are more reliable than code ids due to it's characteristics that it does not change from env to env. That's why I think it's ok to hard code it and have it defined once (in sqs repo). Even though the contract gets migrated and it requires no change on sqs. It will still work. These infos are bound to conversion logic anyway.

For versioning matching, we can use semver matching to check on which pool version we are operating on as well to increase range of compatability. For example, for alloy transmuter it can be ">=3.0.0" for now.

Other cosmwasmpool can also adopt this approach (I didn't update then in this PR since I would love to push this out earlier to test it with USDT alloyed, it requires a bit more regression testing).

Bottom line is that, we can't avoid code changes if the logics for calculating token out changes. We can be lax on the config (eg. ">=3.0.0" and no restriction on the upper bound) until there is a version that requires new calculation. We can have sqs disregard newer versions via config if that's a desired property of the system to control incompatible pool type to not get calculated.

On another point, "matching name only" does not work well IMO. Since there might be changes in the pool transformation logic in the future version, which means we have to change contract name again.

config.json
iboss-ptk: I can't access `infrastructure` repo

domain/router.go
iboss-ptk: done

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
iboss-ptk: done

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
iboss-ptk: done

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
iboss-ptk: done

router/usecase/pools/routable_cw_alloy_transmuter_pool_test.go
iboss-ptk: done

sqsdomain/cw_pool.go
p0mvn: For visibility, this is Boss's reply (did not get submitted publicly) that I will address in a separate message

image

p0mvn: I think that we should discuss this decision via a separate ADR to avoid blocking this line of work.

I would like to avoid merging partial progress with a new paradigm that has trade-offs. If we want to change how CW pools are handled, let's aim to change this for all pools at once rather than introducing custom handling for the alloyed pool but keeping everything else via code IDs.

What I would like to avoid is us never coming back to this and then having to deal with operational challenges stemming from the custom logic.

In terms of versions, I don't understand why we need to track these hard-coded versions in general when contract names should suffice. We should establish a rule that all future versions of a cw pool model struct transferred between node and sqs are backward compatible. This removes the need for version matching on the node side, removing the need to either have a config (node-side) or hard code it in `sqsdomain`.

I'm supportive of using versions in the SQS config rather than hardcoding. But again, I propose making that decision in a separate ADR and/or issue to avoid overly long debates on the otherwise complete work here.

Again, my main counter-argument to the proposed change in this PR is the following:
- Operational challenges stemming from hardcoding versions on both osmosis and sqs side
- Applying an operational change to one cw pool only, causing inconsistencies in handling
- The trade-offs are not fully analyzed before making a change that impacts service operations

However, I think that we should explore the version idea in SQS configs in general. That is an interesting proposal.

@iboss-ptk please let me know what you think

iboss-ptk: Fair point on consistency.

I will revert back the config to code ids and and have a separated ADR for versioning we discussed to unblock things. That being said. Since old transmuter is not forward compatible, it still need minimum version `>=3.0.0` hardcoded. Because even though it is a non-issue on the sqs side in case there is no minimum version in place, ingester needs this information to do pool transformation properly as old transmuter does not have those queries we use for v3.0.0. 

And current transmuter v3 on mainnet is already `crates.io:transmuter` `v3.0.0` so this can't be changed to `crates.io:transmuter-alloy` unless having a migration (and I don't think that make much sense anyway since it is still a newer version of the same thing).


To summarize:
- I don't think we can get rid of any hardcoded version except making more lax and keep backward compatibility. Which might or might not make sense depending on

sqsdomain/cw_pool.go
```diff
@@ -42,7 +51,15 @@ func NewCWPoolModel(contract string, version string, data CWPoolData) *CosmWasmP
...
-	return model.ContractInfo.Matches("crates.io:transmuter", "3.0.0")
+	name := "crates.io:transmuter"
+	version := ">= 3.0.0"
+
+	constraints, err := semver.NewConstraint(version)
+	// this must never panic
+	if err != nil {
+		panic(err)
+	}
+	return model.ContractInfo.Matches(name, constraints)
```
p0mvn, line 62: I like this approach. It is now extensible without having to make code changes to both, the `osmosis` node side and `sqs` for minor migrations.

Can we also add constraint of semver being less than <4.0.0 so that, if there are any future versions, those are auto-accounted for without more code changes?

sqsdomain/cw_pool.go
p0mvn: Thanks

Per [this comment](https://github.com/osmosis-labs/sqs/pull/245#discussion_r1625856060), the latest solution is more extensible.

If we keep it as a convention that all alloyed transmuters stay within the same major version and add a relevant semver constraint (`<4.0.0`), we solve all future extensibility problems.

In the original approach, due to semver being hardcoded directly, we would have to make code changes and dependency updates to all of `sqsdomain`, `osmosis` and `sqs` if a migration was done.

Now, migrations within the same major are auto-accounted for

config-testnet.json
```diff
@@ -27,10 +27,12 @@
             3084,
             4643
         ],
+        "alloyed-transmuter-code-ids": [8319],
         "general-cosmwasm-code-ids": [
             5005,
             6688
-        ]
+        ],
+        "is-alloyed-transmuter-enabled": true
```
p0mvn:
```suggestion
        ]
```

docs/architecture/COSMWASM_POOLS.MD
```diff
@@ -0,0 +1,46 @@
+# CosmWasm Pools
```
p0mvn, line 1: Do you mind adding a paragraph about semver and the general new version handling we arrived at with the cw pools? For example, "alloyed transmuter versions are expected to exist within the same major version i.e >=3.0.0 and <4.0.0"

p0mvn (approved):

config-testnet.json
p0mvn: comitting nit to avoid having you wait on CI

p0mvn:
@iboss-ptk thanks for addressing all operational concerns. Great work.

I pre-approved the PR but the following things remain:
- Update `infrastructure` repository with the new config values: https://github.com/osmosis-labs/infrastructure
  * Please do no merge before this is done because our stage deployments will break
  * You should have access. If still not, please ask in #eng-platform-support
- Add docs about semver handling

sqsdomain/cw_pool.go
iboss-ptk: At least for now that I'm aware of, 4.0.0 will still be compatible, so I would avoid constraining that if you don't want code change. Until there is a version the requires changes in logic, I think leave this lax constraints would be easier to operate.

p0mvn: I think we should have a standard where a separate major means a breaking change.

If `4.0.0` is compatible, can it be `3.1.0` instead?

iboss-ptk: It's breaking other parts. For 4.0.0 we are planning to remove some entrypoints, putting more constraints on denom, but just does not effect sqs.

p0mvn: Got it, let's leave it as is. Thank you!

PR #248, opened 2024-05-23
chore/test: reduce flakiness in prices tests

p0mvn:
New prices tests started being flaky.

The reason was that for some low vol/liq tokens, we would be missing a coingecko ID.

As a result, I refactored to only run the CoinGecko check for the mid and high-volume and liquidity tokens.

As an outcome, I moved the CoinGecko logic towards the end of the test and cleaned up some prints

### DIFF
--- a/tests/test_tokens_prices.py
+++ b/tests/test_tokens_prices.py
@@ -53,22 +53,22 @@ class TestTokensPrices:
     # NUM_TOKENS_DEFAULT low liquidity tokens
     @pytest.mark.parametrize("token",setup.choose_tokens_liq_range(NUM_TOKENS_DEFAULT, MIN_LIQ_FILTER_DEFAULT, MAX_VAL_LOW_LIQ_FILTER_DEFAULT))
     def test_low_liq_token_prices(self, environment_url, token):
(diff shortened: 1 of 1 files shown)


tests/test_tokens_prices.py
```diff
@@ -120,37 +120,36 @@ def test_top_volume_token_prices_in_batch(self, environment_url):
...
-        if (not allow_blank_coingecko_id):
-            assert coingecko_id is not None, f"{token} coingecko id is none"
-        else:
-            return
-
-        # Assert coingecko price is available for the token
-        coingecko_price = coingecko_service.get_token_price(coingecko_id)
-        assert coingecko_price is not None, f"{token},{coingecko_id} coingecko price is none"
-        print(f"{datetime.fromtimestamp(time.time()).strftime(date_format)}: {token},{coingecko_id}, coingecko price = {coingecko_price}")
-        assert coingecko_price > 0, f"{token},{coingecko_id} coingecko price is zero"
```
p0mvn, line 136: Note to reviewer: I simply moved this down since this is an optional check

cryptomatictrader (approved):
All good. Thanks!

PR #255, opened 2024-05-25
docs: release doc basics

p0mvn:
This documents SQS release process and rules around versioning

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,7 +12,12 @@ This is a summary of the steps to get a change into production.
 6. Test & request QA
 7. Tag the non-RC release e.g. `v0.20.0` and push it to the repo
 8. Repeat step 3 and 4 for prod
-9. Manually perform prod deployment
+  * Reference the configuration from step 4.
+9.  Manually perform prod deployment via Rundeck
+  * Post updates in #eng-team-data-services.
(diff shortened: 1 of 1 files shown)


cryptomatictrader (commented):
Thanks for summarizing it in the doc!

One thing I'm unsure is: what is the criteria for incrementing the rc number only?

p0mvn:
> Thanks for summarizing it in the doc!
> 
> One thing I'm unsure is: what is the criteria for incrementing the rc number only?

Good point.

The rc numbers should be incremented for any changes that occur in staging and before getting to production.

Let's say that we plan a release, and the current version is `v0.18.5`. Then, we tag a `v0.18.-6-rc0` and push it to the stage for testing. QA uncovers a bug during stage testing. We push a fix, incrementing the rc number for a new stage tag - `v0.18.6-rc1`. Now, the stage QA passes, and we proceed to production with `v0.18.6`.

I will update the document with this context. Let me know what you think

p0mvn:
@cryptomatictrader updated docs. This is ready for a final review/ACK

cryptomatictrader:
Sorry for the review delay. Looks good to me! I'll also fill it in if I run into anything additional. Thanks!

PR #257, opened 2024-05-25
refactor: pricing worker

p0mvn:
This change is partial progress towards: https://github.com/osmosis-labs/sqs/pull/224

and is extracted for review simplicity.

When pricing worker was first implemented, there was a synchronization mechanism added to prevent concurrent price updates. However, it proved to be unnecessary with further tests.

At the same time, it was limiting the pipeline of workers, preventing the future liquidity pricing worker from executing in-sequence due to the aforementioned synchronisation mechanism.

This PR removes the synchronization mechanism to allow for future extension in #224


cryptomatictrader (approved):
All look good to me. It's become much simpler after removing the synchronization mechanism! Thanks!

PR #261, opened 2024-05-27
POC: Docker debug config

p0mvn:
This PR adds a VS Code configuration for jumping onto our stage/prod nodes via SSH and attaching a debugger to a running Docker container.

This solution requires maintaining a separate Dockerfile for building a debug container with dlv server listening


cryptomatictrader (approved):
Looks good to me. Thanks for making this!

PR #269, opened 2024-05-29
fix: Astroport PCL spot price issues in quotes using quote estimation method

p0mvn:
## Context

Astroport PCL spot price method is broken.

We previously attempted to hot fix it via https://github.com/osmosis-labs/sqs/pull/247 where we would use a quote estimation method by swapping 10^6 of quote denom in.

While that fixed some issues, there were still open ends due to the amount in being static.

This PR introduces a dynamic method for choosing amounts to swap in by using the denom scaling factor.

## Testing

- The new abstractions are validated via table-driven tests with mocks.
- We will use e2e test suite to further validate the quality of the new fallback system for PCL pools


domain/pools.go
```diff
@@ -9,3 +16,40 @@ type CosmWasmPoolRouterConfig struct {
...
+	Calculate(ctx context.Context, baseDenom string, quoteDenom string) (osmomath.BigDec, error)
+}
+
+// UnsetScalingFactorGetterCb is a callback that is used to unset the scaling factor getter callback.
+var UnsetScalingFactorGetterCb ScalingFactorGetterCb = func(denom string) (osmomath.Dec, error) {
+	// Note: for many tests the scaling factor getter cb is irrelevant.
+	// As a result, we unset it for simplicity.
+	// If you run into this panic, your test might benefit from properly wiring the scaling factor
+	// getter callback (defined on the tokens use case)
+	panic("scaling factor getter cb is unser")
```
cryptomatictrader: unser => unset

router/usecase/pools/spot_price_quote_calculator_test.go
```diff
@@ -0,0 +1,230 @@
...
+// the spot price via quote as intended and returns errors in edge cases
+// as given by the spec.
+//
+// To setup the environment, it configures 2 mocks:
+// 1. Mock quote estimator callback
+// 2. Scaling factor computer callback.
+//
+// Then, by using the mocked outputs, it validates the behavior to be as intended.
+// For details about mock configuration, read their respective setup functions.
+func (s *RoutablePoolTestSuite) TestSpotPriceQuoteCalculator_Calculate() {
```
cryptomatictrader, line 26: The mock setup below looks a bit bulky. Does it make sense to move the mock things to domain/mocks ?

cryptomatictrader (commented):
Great work 👍 Just some minor comments

domain/pools.go
p0mvn: Updated

router/usecase/pools/spot_price_quote_calculator_test.go
p0mvn: I would agree if these mocks were meant to be reused across the system.

However, they are intended to be specific for this test.

As a result, to avoid polluting the test-global scope, I propose keeping them here despite the bulkiness

cryptomatictrader (approved):
Good to go!

PR #270, opened 2024-05-30
fix/perf: optimize splits by caching expensive repeated calls

p0mvn:
While working on https://github.com/osmosis-labs/sqs/pull/268 (PEPE splits bug), I noticed that a performance optimization from before was not functioning correctly due to misues of closures.

The performance optimization is intended to cache repeated quote estimation calls in the dynamic programming solution within closure scope to avoid expensive recomputation

This PR fixes the bug, and adds a sanity check test and benchmark.

According to the benchmark, this introduces an 80% speed up to splits and reduces reallocations by the same magnitude.

```
goos: darwin
goarch: arm64
pkg: github.com/osmosis-labs/sqs/router/usecase
                 │  bench_old  │              bench_new              │
                 │   sec/op    │   sec/op     vs base                │
GetSplitQuote-11   814.4µ ± 2%   159.0µ ± 1%  -80.47% (p=0.000 n=10)

                 │  bench_old   │              bench_new               │
                 │     B/op     │     B/op      vs base                │
GetSplitQuote-11   817.6Ki ± 0%   155.1Ki ± 0%  -81.03% (p=0.000 n=10)

                 │  bench_old   │              bench_new              │
                 │  allocs/op   │  allocs/op   vs base                │
GetSplitQuote-11   22.724k ± 0%   4.336k ± 0%  -80.92% (p=0.000 n=10)
```

## Drive By

I also regenerated mainnet test state


tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
```diff
@@ -109,7 +109,7 @@ func (s *CoingeckoPricingTestSuite) TestGetPrices_Coingecko_FindUnsupportedToken
 
 	fmt.Println("Total unsupported tokens: ", unsupportedCounter)
 
-	// Total unsupported tokens as of May 24 2024: 136
-	s.Require().Equal(136, unsupportedCounter)
+	// Total unsupported tokens as of May 29 2024: 151
+	s.Require().Equal(151, unsupportedCounter)
```
p0mvn, line 113: This is a sharp increase but should not be affected by the changed logic here.

This probably stems from mainnet state updates / asset listings.

Will run the e2e test suite on this prior to merge

router/usecase/dynamic_splits.go
```diff
@@ -173,6 +139,7 @@ func getSplitQuote(ctx context.Context, routes []route.RouteImpl, tokenIn sdk.Co
 		isAmountInNilOrZero := inAmount.IsNil() || inAmount.IsZero()
 		isAmountOutNilOrZero := outAmount.IsNil() || outAmount.IsZero()
 		if isAmountInNilOrZero && isAmountOutNilOrZero {
+			// totalIncrementsInSplits += currentRouteIncrement
```
cryptomatictrader: I guess there is no need to keep this line anymore?

cryptomatictrader:
Huge improvement made 💯 Thanks!

cryptomatictrader (approved):

tokens/usecase/pricing/coingecko/pricing_coingecko_test.go
p0mvn: I have checked this by only copying the test state update and observing the same number of breakages:
https://github.com/osmosis-labs/sqs/pull/273

router/usecase/dynamic_splits.go
p0mvn: Removed

PR #278, opened 2024-06-03
[Dynamic Min Liquidity] implement denom pool liquidity data transformations

p0mvn:
Ref: https://linear.app/osmosis/issue/DATA-199/[dynamic-min-liquidity-filter]-implement-and-test-new-ingester-data

Separated from https://github.com/osmosis-labs/sqs/pull/224 for ease of review.

To compute dynamic min liquidity based on the liquidity of each denom across all pools, we must transform the pool data into a mapping between denoms and their corresponding liquidity.

This PR implements the necessary helpers infrastructure to achieve this, propagating the data into a pricing worker for further usage.

See #224 for the end result of this feature.

## Testing

- #224 has already been tested on node
- The new logic in this PR is covered by unit tests

### DIFF
--- a/domain/mocks/pricing_update_listener_mock.go
+++ b/domain/mocks/pricing_update_listener_mock.go
@@ -4,12 +4,13 @@ import (
 	"context"
 	"time"
 
+	"github.com/osmosis-labs/osmosis/osmomath"
 	"github.com/osmosis-labs/sqs/domain"
 )
 
 type PricingUpdateListenerMock struct {
 	Height                 int
-	PricesBaseQuteDenomMap map[string]map[string]any
+	PricesBaseQuteDenomMap map[string]map[string]osmomath.BigDec
 	QuoteDenom             string
 
 	Done chan struct{}
@@ -22,7 +23,7 @@ type PricingUpdateListenerMock struct {
 func NewPricingListenerMock(timeout time.Duration) *PricingUpdateListenerMock {
 	return &PricingUpdateListenerMock{
 		Done:                   make(chan struct{}),
-		PricesBaseQuteDenomMap: make(map[string]map[string]any),
+		PricesBaseQuteDenomMap: make(map[string]map[string]osmomath.BigDec),
 		timeout:                timeout,
 	}
 }
@@ -30,7 +31,7 @@ func NewPricingListenerMock(timeout time.Duration) *PricingUpdateListenerMock {
 var _ domain.PricingUpdateListener = &PricingUpdateListenerMock{}
 
 // OnPricingUpdate implements worker.PricingUpdateListener.
-func (p *PricingUpdateListenerMock) OnPricingUpdate(ctx context.Context, height int64, pricesBaseQuoteDenomMap map[string]map[string]any, quoteDenom string) error {
+func (p *PricingUpdateListenerMock) OnPricingUpdate(ctx context.Context, height uint64, blockMetadata domain.BlockPoolMetadata, pricesBaseQuoteDenomMap domain.PricesResult, quoteDenom string) error {
 	p.Height = int(height)
 	p.QuoteDenom = quoteDenom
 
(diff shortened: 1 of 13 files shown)


domain/mocks/pricing_update_listener_mock.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"time"
 
+	"github.com/osmosis-labs/osmosis/osmomath"
 	"github.com/osmosis-labs/sqs/domain"
 )
 
 type PricingUpdateListenerMock struct {
 	Height                 int
-	PricesBaseQuteDenomMap map[string]map[string]any
+	PricesBaseQuteDenomMap map[string]map[string]osmomath.BigDec
```
cryptomatictrader, line 13: Just question not comment - guess we have to use osmomath.BigDec as it involves price precision, right? I remember you said excessive use of osmomath.BigDec can have performance impact

domain/mvc/tokens.go
```diff
@@ -34,7 +34,7 @@ type TokensUsecase interface {
 	// The outer map consists of base denoms as keys.
 	// The inner map consists of quote denoms as keys.
 	// The result of the inner map is prices of the outer base and inner quote.
-	GetPrices(ctx context.Context, baseDenoms []string, quoteDenoms []string, pricingSourceType domain.PricingSourceType, opts ...domain.PricingOption) (map[string]map[string]any, error)
+	GetPrices(ctx context.Context, baseDenoms []string, quoteDenoms []string, pricingSourceType domain.PricingSourceType, opts ...domain.PricingOption) (domain.PricesResult, error)
```
cryptomatictrader, line 37: 👍 Much better than map of string to a map of string to any

domain/pricing.go
```diff
@@ -142,6 +142,15 @@ type PricingWorker interface {
...
+	// OnPricingUpdate notifies the listener of the pricing update.
+	OnPricingUpdate(ctx context.Context, height uint64, blockMetaData BlockPoolMetadata, pricesBaseQuoteDenomMap PricesResult, quoteDenom string) error
 }
+
+// PricesResult defines the result of the prices.
+// [base denom][quote denom] => price
+// Note: BREAKING API - this type is API breaking as it is serialized to JSON.
+// from the /tokens/prices endpoint. Be mindful of changing it without
+// separating the API response for backward compatibility.
+type PricesResult map[string]map[string]osmomath.BigDec
```
cryptomatictrader, line 156: See the map of string to a map of string to BigDed here. It'll be good if we can have an example of the structure here.

domain/mocks/pricing_update_listener_mock.go
p0mvn: Yes, it does have a performance impact since it is 36 decimals as opposed to 18 decimals with regular Dec.

We allocate raw bytes to represent this structure under the hood. The performance cost is most notable in intermediary math operations since we now need to operate on more bytes.

The native ints are usually preferred for performance, if they are sufficient to represent what we need. 

With the on-chain price, however, we need 36 decimals to reduce the likelihood of truncations in intermediary operations since prices can either be large or small due to a potential for orders of magnitude in difference in reserves. Additionally, decimal precision exponents exacerbate this. As a result, using `BigDec` for prices at the expense of performance is an acceptable performance cost.

This is also not the performance bottleneck today

ingest/usecase/ingest_usecase_test.go
```diff
@@ -1 +1,369 @@
...
+	"github.com/osmosis-labs/sqs/router/usecase/routertesting"
+	"github.com/stretchr/testify/suite"
+)
+
+type IngestUseCaseTestSuite struct {
+	routertesting.RouterTestHelper
+}
+
+const (
+	defaltPoolID uint64 = 1
```
cryptomatictrader: Typo in the variable

cryptomatictrader (approved):
LGTM 👍

ingest/usecase/ingest_usecase_test.go
p0mvn: Fixed

Issue #132, opened 2024-03-18
[SQS-Routing] API to force quote over a given route


p0mvn:
We have frequent requests claiming that another route might be more performant without a streamlined way to confirm it.

To make the discussions more productive and to avoid investigating the details of route differences, we should have a Swagger API that allows to force quote from token A to token B over given pool IDs / routes.

## Acceptance Criteria

* Add an API to achieve this
* Test it
* Expose via swagger
* Deploy to stage
* Deploy to production
* Add add alerts if applicable
* Add Grafana dashboards if applicable 
* Backwards compatibility with `custom-direct-quote` . Simply extending it to support more pools

p0mvn:
@deividaspetraitis would you be interested in this one?

Instead of creating a new endpoint, we can modify the existing one:
https://github.com/osmosis-labs/sqs/blob/a812e0f87620b114c621afbbfb8575d3ab8730ab/router/delivery/http/router_handler.go#L47

However, we would need to ensure backwards compatibility

deividaspetraitis:
Sure! Please feel free to assign me and let me take a close look once I'm back to my PC, I think I might have a few questions.

deividaspetraitis:
hey @p0mvn, I just wanted to clarify a few details to make sure I understand task and scope correctly. 

As per my understanding the idea is to have a Swagger API that allows to force quote from token A to token B over given pool IDs / routes, which is basically the functionally of  `router/custom-direct-quote` endpoint as you have pointed. Is my understanding correct that it's basically following that needs to be implemented ( within SQS code base ):

1. Extend  `router/custom-direct-quote` endpoint to support more pools ( with backwards compatibility ). 
2. Add relevant annotations  for `router/custom-direct-quote` so it's available in Swagger API UI ( [here](https://sqs.osmosis.zone/swagger/index.html#/default/get-token-metadata) ).

In case I am on the right track with above, I was wondering what specific new pools do we need add to support? Is that about [Custom CosmWasm Pools](https://github.com/osmosis-labs/sqs?tab=readme-ov-file#custom-cosmwasm-pools)? I am sorry for my ignorance. :)

p0mvn:
You're pretty much spot on here.

Please note that, currently, we support `poolID` [query param](https://github.com/osmosis-labs/sqs/blob/adafa8dff12dbb044a2d8e5192e9e109bd962bf1/router/delivery/http/router_handler.go#L133-L136).

[...]

PR #284, opened 2024-06-04
[SQS-Routing] API to force quote over a given route

deividaspetraitis:
This pull request implemets improvements for API to force quote over a given ( pools ) route. For more background please see #132.

## Background:

Work is based on the existing `/router/custom-direct-quote` endpoint.

## Implementation notes:

* Instead of adding extra `poolIDs` parameter, it seemed it is simpler to reuse existing `poolID` to avoid complexity and additional code
* `tokenOutDenom` parameter was adjusted to accept not only singe denonm but also a comma separated list of token out denoms
* All changes were made with backwards compatability in mind

Looking forward for reviews and comments!


router/usecase/optimized_routes_test.go
```diff
@@ -674,6 +674,72 @@ func (s *RouterTestSuite) TestGetCustomQuote_GetCustomDirectQuote_Mainnet_UOSMOU
...
+
+		err error
+	}{
+		{
+			name:                "OSMO/USDC - happy case",
+			poolID:              []uint64{1369, 1374, 1484},
+			expectedNumOfRoutes: 3,
+			expectedPoolID:      []uint64{1369, 1374, 1484},
+		},
+		// TODO:
```
deividaspetraitis: For example, pool with ID `1` does not have `OSMO/USDC` pair, what we sholud do?

router/delivery/http/router_handler.go
deividaspetraitis, line 1: TODO: cover with tests?

router/usecase/router_usecase.go
```diff
@@ -448,6 +453,27 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
 	return getBestSingleRouteQuote(ctx, tokenIn, routes, r.logger)
 }
 
+// GetCustomDirectQuoteMultiPool implements mvc.RouterUsecase.
+func (r *routerUseCaseImpl) GetCustomDirectQuoteMultiPool(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom string, poolID []uint64) (domain.Quote, error) {
```
deividaspetraitis: The following is the expected approach?

deividaspetraitis:
Hey @p0mvn, can I ask a moment of your time to take a look at this draft and provide your insights whether I am on the right track delivering #132, please?

router/usecase/optimized_routes_test.go
p0mvn: If you are translating some previous example I gave literally, I made it up. You can construct similar example test cases by checking out the pools.

Pool 1 should be `routertesting.UOSMO` and `routestertesting.ATOM`

See:
https://app.osmosis.zone/pool/1

or

```
osmosisd q poolmanager pool 1 --node https://rpc.osmosis.zone:443
```

router/usecase/optimized_routes_test.go
```diff
@@ -674,6 +674,72 @@ func (s *RouterTestSuite) TestGetCustomQuote_GetCustomDirectQuote_Mainnet_UOSMOU
...
+	routerRepositoryMock.SetTakerFees(mainnetState.TakerFeeMap)
+
+	// Setup pools usecase mock.
+	poolsUsecase := poolsusecase.NewPoolsUsecase(&domain.PoolsConfig{}, "node-uri-placeholder", routerRepositoryMock, domain.UnsetScalingFactorGetterCb)
+	poolsUsecase.StorePools(mainnetState.Pools)
+
+	routerUsecase := routerusecase.NewRouterUsecase(routerRepositoryMock, poolsUsecase, config, emptyCosmWasmPoolsRouterConfig, &log.NoOpLogger{}, cache.New(), cache.New())
+
+	// Test cases
+	testCases := []struct {
```
p0mvn: Suggest also adding simpler cases:
- Empty pool IDs
- One valid pool ID
- One pool ID that does not contain one of the tokens

router/usecase/router_usecase.go
```diff
@@ -448,6 +453,27 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
...
+	for _, v := range poolID {
+		quote, err := r.GetCustomDirectQuote(ctx, tokenIn, tokenOutDenom, v)
+		if err != nil {
+			return nil, err
+		}
+
+		i, ok := quote.(*quoteImpl)
+		if !ok {
+			return nil, fmt.Errorf("quote ( v=%v ) is not *quoteImpl", quote)
+		}
```
p0mvn: nit: I don't have any specific recommendations but could you check if there is a way we could avoid casting?

If non-trivial, feel free to resolve

router/usecase/router_usecase.go
```diff
@@ -448,6 +453,27 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
 	return getBestSingleRouteQuote(ctx, tokenIn, routes, r.logger)
 }
 
+// GetCustomDirectQuoteMultiPool implements mvc.RouterUsecase.
+func (r *routerUseCaseImpl) GetCustomDirectQuoteMultiPool(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom string, poolID []uint64) (domain.Quote, error) {
+	var result quoteImpl
+	for _, v := range poolID {
+		quote, err := r.GetCustomDirectQuote(ctx, tokenIn, tokenOutDenom, v)
```
p0mvn, line 464: As we do the iteration, `tokenIn` and `tokenOutDenom` should change.

The token out amount of the previous iteration becomes the token in of the current.
The token out denom of the previous iteration becomes the token in of the current.

I believe that I've missed a requirement that we also need to provide a sequence of token out denoms.

Note that we have a chain message with a similar API [here](https://github.com/osmosis-labs/osmosis/blob/0f01689ae77b47e74bb72fc4f3258f8c470c18fc/x/poolmanager/router.go#L53-L58).

This is [the structure](https://github.com/osmosis-labs/osmosis/blob/0f01689ae77b47e74bb72fc4f3258f8c470c18fc/x/poolmanager/types/swap_route.pb.go#L27-L30) that contains a pool ID and token out denom per iteration step that is supposed to be given by the user.

deividaspetraitis: Aha! I seems I had a bit different understanding what needs to be done. Please see updated logic for this method and let me know whether it's something that you're expecting ( in a high level ). 
Note that I was not able to fully reuse `GetCustomDirectQuote`, just some parts of it. It is because if a pair of asset let's say `OSMO/USDC` is considered to be swapped through multiple pools this [method will error](https://github.com/deividaspetraitis/sqs/blob/d3483090075258e2dbec3127a775319893a6b636/router/usecase/router_usecase.go#L415) because out denom is going to be something else at that stage but not `USDC` until the final pool.

P.S. I am aware that I need to include calculating taker fee which is missing at this point inculding further cleanup.

router/usecase/router_usecase.go
deividaspetraitis: I have updated as per your suggestion.

router/usecase/optimized_routes_test.go
```diff
@@ -674,6 +674,118 @@ func (s *RouterTestSuite) TestGetCustomQuote_GetCustomDirectQuote_Mainnet_UOSMOU
...
+			name:          "Multi pool: OSMO-USDC - happy case",
+			tokenIn:       sdk.NewCoin(UOSMO, amountIn),
+			tokenOutDenom: USDC,
+			poolID: []uint64{
+				1093, // OSMO - AKT
+				1480, // AKT - USDC
+			},
+			expectedNumOfRoutes: 2,
+			expectedPoolID:      []uint64{1093, 1480},
+		},
```
p0mvn: Let me construct a made up example that will currently break

```go
tokenIn:       sdk.NewCoin(UOSMO, amountIn),
tokenOutDenom: USDC,
poolID: []uint64{
   5, // OSMO, AKT, ATOM
   24 // AKT - USDC
}
```

In this example, when we complete the iteration over the first pool, how do we know which tokenOutDenom becomes the tokenIn denom for the next iteration?

We need to add another query param `tokenOutDenoms` which is a slice of token out denoms corresponding to the pool ID stemming from `poolIDs` query param in the same position.

Just for context (we should not pursue this) - alternatively, we could try using `GetCandidateRoutes` route that would construct the slice of pool IDs and respective tokenOutDenoms on the path from the user-provided token in denom to the user-provided token out denom.

However, the way this algorithm works is that it finds 20 routes between two tokens and does not search further. In some cases, we might want to investigate routes that this routine cannot find. As a result, we should push the responsibility of selecting the pool IDs and relevant token out denoms to the user.

What I'm thinking about is something like this:
```go
func GetCustomDirectQuoteMultiPool(ctx context.Context, tokenIn sdk.Coin, poolID []uint64, tokenOutDenoms []string) {
   for i, poolID := range poolIDs {
      curTokenOutDenom :=  tokenOutDenoms[i]
      
      // do stuff
      curAmtOut, ... := GetCustomDirectQuote(..., tokenIn, curTokenOutDenom)
     
      tokenIn = sdk.NewCoin(curTokenOutDenom, curAmtOut)
   }
}
```

Please let me know if that makes sense ⏫

router/delivery/http/router_handler.go
```diff
@@ -144,20 +148,16 @@ func (a *RouterHandler) GetDirectCustomQuote(c echo.Context) error {
...
-	quote, err := a.RUsecase.GetCustomDirectQuote(ctx, tokenIn, tokenOutDenom, poolID)
+	quote, err := a.RUsecase.GetCustomDirectQuoteMultiPool(ctx, tokenIn, tokenOutDenom, poolIDs)
 	if err != nil {
 		return c.JSON(domain.GetStatusCode(err), domain.ResponseError{Message: err.Error()})
 	}
 
 	scalingFactor := oneDec
 	if shouldApplyExponents {
-		scalingFactor = a.getSpotPriceScalingFactor(tokenIn.Denom, tokenOutDenom)
+		// TODO: is this right approach to take last token out?
```
deividaspetraitis: Is this right approach to calculate spot price scaling factor?

router/usecase/optimized_routes_test.go
deividaspetraitis: I have updated tests to include such test case. Please feel free to resolve if that's something that you would expect.

router/usecase/optimized_routes_test.go
deividaspetraitis: That totally makes sense and explains everything, thanks for taking your time! I was asking the very same question myself while working on the task since your last [comment](https://github.com/osmosis-labs/sqs/pull/284#discussion_r1628725802), but somehow your suggestion there slipped through my eyes which led previous version of the solution. Marking this thread as resolved for now.

router/usecase/optimized_routes_test.go
deividaspetraitis: Marking as resolved, list of out denom tokens aswered to this question.

router/delivery/http/router_handler.go
deividaspetraitis: Marking as resolved for now, I think it already took a bit more time than I was expecting initially for this task. Nevertheless, should we strive to implement incrementally handler tests over next PR?

tokens/delivery/http/tokens_delivery.go
```diff
@@ -130,7 +130,7 @@ func (a *TokensHandler) GetMetadata(c echo.Context) (err error) {
 // @Produce  json
 // @Param   base          query     string  true  "Comma-separated list of base denominations (human-readable or chain format based on humanDenoms parameter)"
 // @Param   humanDenoms   query     bool    false "Specify true if input denominations are in human-readable format; defaults to false"
-// @Param	pricingSource query		int     false "Specify the pricing source. Values can be 0 (chain) or 1 (coingecko); default to 0 (chain)
+// @Param	pricingSource query     int     false "Specify the pricing source. Values can be 0 (chain) or 1 (coingecko); default to 0 (chain)"
```
deividaspetraitis, line 133: Quick fix to make swagger happy and do not complain about parsing errors.

router/usecase/router_usecase.go
```diff
@@ -448,6 +439,50 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
...
+	var result quoteImpl
+	for i, v := range poolID {
+		tokenOutDenom := tokenOutDenom[i]
+
+		quote, err := r.GetCustomDirectQuote(ctx, tokenIn, tokenOutDenom, v)
+		if err != nil {
+			return nil, err
+		}
+
+		// fill the amount in token with the first asset of the pair
```
deividaspetraitis: Are those `AmountIn` and `AmountOut` assignments correct?

deividaspetraitis:
hey @p0mvn, please see updated PR as per your suggestions, I am marking this PR ready as review because I think we are close resolving it, some information was added in the PR description to make it more clear and more informative. I have started a several threads for the places in the code where I felt your input would help to clarify implementation correctness and closed older ones to reduce noise and make this PR more cleaner.

router/usecase/router_usecase.go
```diff
@@ -448,6 +439,50 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
 	return getBestSingleRouteQuote(ctx, tokenIn, routes, r.logger)
 }
 
+var ErrValidationFailed = fmt.Errorf("validation failed")
+
+// GetCustomDirectQuoteMultiPool implements mvc.RouterUsecase.
+func (r *routerUseCaseImpl) GetCustomDirectQuoteMultiPool(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom []string, poolID []uint64) (domain.Quote, error) {
```
p0mvn: nit

```suggestion
func (r *routerUseCaseImpl) GetCustomDirectQuoteMultiPool(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom []string, poolIDs []uint64) (domain.Quote, error) {
```

router/usecase/router_usecase.go
```diff
@@ -448,6 +439,50 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
...
+
+		quote, err := r.GetCustomDirectQuote(ctx, tokenIn, tokenOutDenom, v)
+		if err != nil {
+			return nil, err
+		}
+
+		// fill the amount in token with the first asset of the pair
+		if i == 0 {
+			result.AmountIn = tokenIn
+		}
```
p0mvn: Could we instead initialize `result` on line 459 to this value?

router/usecase/router_usecase.go
```diff
@@ -448,6 +439,50 @@ func (r *routerUseCaseImpl) GetCustomDirectQuote(ctx context.Context, tokenIn sd
...
+
+		// fill the amount of the second asset of the pair
+		if i == len(poolID)-1 {
+			result.AmountOut = quote.GetAmountOut()
+		}
```
p0mvn: To reduce cognitive overhead, can we update the `result.AmountOut` on every iteration?

router/delivery/http/router_handler.go
```diff
@@ -346,6 +346,21 @@ func getValidRoutingParameters(c echo.Context) (string, sdk.Coin, error) {
 	return tokenOutStr, tokenIn, nil
 }
 
+// getDirectCustomQuoteParameters returns the pool IDs, tokenIn and tokenOutDenom from server context if they are valid.
+func getDirectCustomQuoteParameters(c echo.Context) ([]uint64, []string, sdk.Coin, error) {
+	poolID, tokenOutStr, tokenInStr, err := getPoolsValidTokenInTokensOut(c)
+	if err != nil {
+		return nil, nil, sdk.Coin{}, err
+	}
```
p0mvn, line 354: Why do we need this new function?

Could we use a combination of the old`getValidRoutingParameters` + the new `getValidPoolID`?

router/usecase/optimized_routes_test.go
```diff
@@ -674,6 +674,143 @@ func (s *RouterTestSuite) TestGetCustomQuote_GetCustomDirectQuote_Mainnet_UOSMOU
 	s.validateExpectedPoolIDOneRouteOneHopQuote(quote, expectedPoolID)
 }
 
+// TODO
+func (s *RouterTestSuite) TestGetCustomQuote_GetCustomDirectQuotes_Mainnet_UOSMOUSDC() {
```
p0mvn: Let's move this into `router_usecase_test.go`? I think we will separately need to move other tests later as well.

The current structure doesn't make much sense

router/usecase/router_usecase.go
p0mvn: LGTM. Left a few nits though

router/delivery/http/router_handler.go
```diff
@@ -380,10 +395,60 @@ func getValidTokenInTokenOutStr(c echo.Context) (tokenOutStr, tokenInStr string,
...
+		if err != nil {
+			return nil, err
+		}
+		poolIDs = append(poolIDs, i)
+	}
+
+	return poolIDs, nil
+}
+
+func getPoolsValidTokenInTokensOut(c echo.Context) (poolID []uint64, tokenOut []string, tokenIn string, err error) {
```
p0mvn: Could we add some unit tests for this?

I would like to make sure that we maintain backward compatibility. Detailed unit tests would help guarantee this and also make future changes simpler

deividaspetraitis: Tests were added, please let me know if you would like to include some additional test case.

router/usecase/optimized_routes_test.go
deividaspetraitis: Moved `TestGetCustomQuote_GetCustomDirectQuotes_Mainnet_UOSMOUSDC` test case to `router_usecase_test.go`. Marking this thread as resolved.

deividaspetraitis:
Thank you for review, @p0mvn! I have updated PR to include suggested changes and also I have added entry in `CHANGLOG.md`.

router/delivery/http/router_handler.go
```diff
@@ -136,21 +144,24 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
+// @Summary Compute the quote for the given poolID
+// @Description Call does not search for the route rather directly computes the quote for the given poolID.
+// @ID get-direct-quote
+// @Produce  json
+// @Param  tokenIn         query  string  true  "String representation of the sdk.Coin for the token in."                   example(5OSMO)
+// @Param  tokenOutDenom   query  string  true  "String representing the list of the token denom out separated by comma."   example(ATOM,USDC)
+// @Param  poolID          query  string  true  "String representing list of the pool ID."                                  example(1,2,3)
+// @Param  applyExponents  query  bool    false  "Boolean flag indicating whether to apply exponents to the spot price. False by default."
+// @Success 200  {object}  domain.Quote  "The computed best route quote"
+// @Router /router/custom-direct-quote [get]
```
p0mvn, line 156: Do you mind please adding humanDenoms support? Please see `/router/quote` for reference

router/delivery/http/router_handler.go
```diff
@@ -136,21 +144,24 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
 func (a *RouterHandler) GetDirectCustomQuote(c echo.Context) error {
```
p0mvn, line 157: We should cover this endpoint by our new integration test suite.

Please see reference here (similar to `/router/quote`:
https://github.com/osmosis-labs/sqs/blob/d63e8c937378c68704976de86cf2d59cdfd6a0e0/tests/test_router_quote.py#L31

Please see this for set up:
https://github.com/osmosis-labs/sqs/blob/d63e8c937378c68704976de86cf2d59cdfd6a0e0/Makefile#L190-L202

p0mvn (approved):
Tested this on a node! Great work @deividaspetraitis 

There are 2 follow-up items that I think we should tackle. Do you mind please helping with these tasks?
- https://github.com/osmosis-labs/sqs/issues/324
- https://github.com/osmosis-labs/sqs/issues/325

#325 is more important so that we can deploy this to production with more confidence that there are no breakages.

Thanks!

PR #291, opened 2024-06-07
Add initial passthrough use case with account assets total endpoint

jonator:
Adds a new usecase called "passthrough" that is merely a client proxy from the SQS client to the node. Repeated node queries are maximally performant within this proxy due to the colocation of the query node + SQS; all networking is on localhost.

The first function of the use new case is `GetPortfolioAssets`, which accepts a valid bech32 address and returns all coins from bank, staking, locks, CL positions, and the underlying coins of any GAMM shares within.

It performs queries and result aggregation concurrently, guaranteeing optimal performance.

It provides a "best-effort" result. If one of the queries or aggregation operations fails, it is intended to skip that result but return all others rather than failing the entire query.

Original [ADR](https://www.notion.so/osmosiszone/ADR-014-Minimize-Latency-for-Geo-Distributed-Queries-a9d85415867941daa3436d3070a2a9b1)

## Swagger API Docs

```
// @Summary Returns portfolio assets associated with the given address.
// @Description The returned data represents the total value of the assets in the portfolio. Total value cap represents the total value of the assets in the portfolio.
// includes capitalization of user balances, value in locks, bonding or unbonding
// as well as the concentrated positions.
// Account coins result represents coins only from user balances (contrary to the total value cap).
// @Produce  json
// @Success 200  struct  passthroughdomain.PortfolioAssetsResult  "Portfolio assets from user balances and capitalization of the entire account value"
// @Router /portfolio-assets/{address} [get]
```

## Performance

image

image

image

## Testng

- Unit tests added for all new components. Low-level components tested as table-driven. Higher-level as happy path
- Load tested
- Integration test added 

## Observability

- Added error logs to highlight unexpected failure paths
- Will add counter metrics and alerts post-merge and stage deployment

## Summary by CodeRabbit

- **New Features**
	- Introduced a passthrough functionality allowing the server to handle passthrough queries efficiently.
	- Added a new API endpoint to retrieve portfolio assets based on a wallet address.
	- Implemented a new HTTP handler for managing passthrough use cases.
	- Enhanced the pricing strategy with new liquidity pricer capabilities.
  
- **Bug Fixes**
	- Streamlined the logic by removing redundant code related to quote denomination retrieval.
  
- **Documentation**
	- Updated API documentation to include the new portfolio assets endpoint.
	- Added detailed descriptions in Swagger files for improved clarity and usability.


passthrough/delivery/http/passthrough_handler.go
```diff
@@ -0,0 +1,46 @@
...
+	handler := &PassthroughHandler{
+		PUsecase: ptu,
+	}
+
+	e.GET(formatPoolsResource("/account_coins_total/:address"), handler.GetAccountCoinsTotal)
+}
+
+// GetAccountCoinsTotal adds an API handler to get total assets data
+func (a *PassthroughHandler) GetAccountCoinsTotal(c echo.Context) error {
+	address := c.Param("address")
```
p0mvn, line 45: nit: let's do some basic param validation - i.e address string is not empty

passthrough/usecase/passthrough_usecase.go
```diff
@@ -0,0 +1,95 @@
...
+	grpcClient, err := grpc.Dial(grpcURI, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, err
+	}
+	
+	return &passthroughUseCase{
+		poolsUseCase: puc,
+
+		bankQueryClient: banktypes.NewQueryClient(grpcClient),
+		stakingQueryClient: staking.NewQueryClient(grpcClient),
```
p0mvn: Suggest initializing query clients in the caller of `NewPassThroughUsecase`. This is so that we can then create mocks for these query clients for unit testing them

passthrough/usecase/passthrough_usecase.go
```diff
@@ -0,0 +1,95 @@
...
+
+			exitCoins, err := p.poolsUseCase.CalcExitCFMMPool(uint64(poolIDInt), balance.Amount)
+			if err != nil {
+				return nil, err
+			}
+			coins = coins.Add(exitCoins...)
+		} else {
+			coins = coins.Add(balance)
+		}
+	}
```
p0mvn: There is a future optimization possible for this.

Currently, we query all balances and filter them out to contain gamm shares.

However, Cosmos SDK has a way to query KV store prefix directly.

As a result, we can query account KV store gamm balance prefix so that this is an efficient database iteration.

This might save us quite a bit of performance.

Consider investigating this once the initial working MVP is ready but suggest not blocking on this comment at this stage.

Let me know if any links to helpful resources on this would help

pools/usecase/pools_usecase.go
```diff
@@ -255,3 +259,24 @@ func (p *poolsUseCase) StorePools(pools []sqsdomain.PoolI) error {
...
 }
+
+func (p *poolsUseCase) CalcExitCFMMPool(poolID uint64, exitingShares osmomath.Int) (sdk.Coins, error) {
+	sqsPool, err := p.GetPool(poolID)
+	if err != nil {
+		return nil, err
+	}
+	
+	if (sqsPool.GetType() != poolmanagertypes.Balancer && sqsPool.GetType() != poolmanagertypes.Stableswap) {
+		return nil, errors.New("invalid pool type, expected CFMM pool")
```
p0mvn: Consider formatting pool ID in error to ease future debugging

passthrough/usecase/passthrough_usecase.go
jonator: Couldn't we create another function like NewMockPassthroughUsecase that does this? This can continue to do the work of setting up actual query clients and encapsulating the actual clients that are used

passthrough/usecase/passthrough_usecase.go
jonator: I'm thinking of supporting all the asset queries, then optimizing them as needed. Thoughts?

jonator:
Any recommendations for how to run queries concurrently instead of serially would be appreciated.

But, from curling in the CLI it seems this query is nearly instant given localhost networking. With that, IMO further optimizations are pre-optimizations and this would accomplish what is needed. Load tests would be a good next step for confirming our performance needs are met as well. Thoughts?

passthrough/usecase/passthrough_usecase.go
p0mvn: We could. However, it's generally preferable to wire up all of the dependencies in the top-level component.

Additionally, fragmenting mock and non-mock constructors would add some cognitive overhead and diverge from the patterns that have been employed in the rest of the repository.

I still think it would make sense to implement this suggestion but LMK what you think

passthrough/usecase/passthrough_usecase.go
p0mvn: SGTM

p0mvn:
> Any recommendations for how to run queries concurrently instead of serially would be appreciated.
> 
> But, from curling in the CLI it seems this query is nearly instant given localhost networking. With that, IMO further optimizations are pre-optimizations and this would accomplish what is needed. Load tests would be a good next step for confirming our performance needs are met as well. Thoughts?


This is a ChatGPT rec that I agree with. I would implement a separate package called `query/concurrent` and dump this in there.

Let me know if this makes sense. I'm happy to jump on the call sometime to discuss various other strategies for concurrency in Go.

Creating an abstraction for executing concurrent queries in Go can be efficiently managed using goroutines and channels. The `sync` package provides tools like `sync.WaitGroup` which can be very helpful in managing concurrency. Here’s a structured approach to achieve this:

### Step-by-Step Guide

1. **Define a Query Function Type:**
   Define a type for your query functions that will be executed concurrently.

   ```go
   type QueryFunc func() (interface{}, error)
   ```

2. **Create a Result Struct:**
   Define a struct to hold the result of each query, including any error.

   ```go
   type QueryResult struct {
       Result interface{}
       Error  error
   }
   ```

3. **Implement the Concurrent Execution Function:**
   Create a function that takes a slice of `QueryFunc` and executes them concurrently.

   ```go
   package main

   import (
       "fmt"
       "sync"
   )

   // QueryFunc is a function type that returns a result and an error
   type QueryFunc func() (interface{}, error)

   // QueryResult holds the result and error of a query
   type QueryResult struct {
       Result interface{}
       Error  error
   }

   // ExecuteConcurrentQueries executes multiple QueryFuncs concurrently
   func ExecuteConcurrentQueries(queries []QueryFunc) []QueryResult {
       var wg sync.WaitGroup
       results := make([]QueryResult, len(queries))
       resultChan := make(chan QueryResult, len(queries))

       // Launch each query in a separate goroutine
       for i, query := range queries {
           wg.Add(1)
           go func(index int, q QueryFunc) {
               defer wg.Done()
               result, err := q()
               resultChan <- QueryResult{Result: result, Error: err}
           }(i, query)
       }

       // Close the result channel once all goroutines are done
       go func() {
           wg.Wait()
           close(resultChan)
       }()

       // Collect the results from the result channel
       for result := range resultChan {
           results = append(results, result)
       }

       return results
   }

   // Example query function
   func exampleQuery(id int) QueryFunc {
       return func() (interface{}, error) {
           // Simulate some work
           return fmt.Sprintf("Result from query %d", id), nil
       }
   }

   func main() {
       queries := []QueryFunc{
           exampleQuery(1),
           exampleQuery(2),
           exampleQuery(3),
       }

       results := ExecuteConcurrentQueries(queries)

       for i, result := range results {
           if result.Error != nil {
               fmt.Printf("Query %d failed: %v\n", i, result.Error)
           } else {
               fmt.Printf("Query %d succeeded: %v\n", i, result.Result)
           }
       }
   }
   ```

### Explanation

- **QueryFunc Type:** A function type that represents the query function. Each query function should return a result and an error.
- **QueryResult Struct:** Holds the result of a query and any error that occurred during its execution.
- **ExecuteConcurrentQueries Function:**
  - Uses a `sync.WaitGroup` to wait for all goroutines to finish.
  - A buffered channel is used to collect results from each goroutine.
  - Each query is executed in a separate goroutine.
  - Results are collected from the channel after all queries are completed.

### Benefits

- **Concurrency:** Leverages Go’s goroutines for concurrent execution, making it efficient.
- **Error Handling:** Captures errors from each query, allowing for robust error handling.
- **Scalability:** Can easily scale with the number of queries by leveraging goroutines.

This approach abstracts the concurrency management, allowing you to focus on the logic of each individual query.

p0mvn:
For load testing, can you add a load test in this file and see how this behaves?

https://github.com/osmosis-labs/sqs/blob/v25.x/locust/locustfile.py#L54

After you add a new load test, this is how you start the UI for it:
https://github.com/osmosis-labs/sqs/blob/267d0e5b17a6de8b867ac40f67b0827b826fe52e/Makefile#L139-L140

Make sure to change the IP on [this line](https://github.com/osmosis-labs/sqs/blob/267d0e5b17a6de8b867ac40f67b0827b826fe52e/locust/docker-compose.yml#L12) to your instance

passthrough/usecase/passthrough_usecase.go
```diff
@@ -0,0 +1,140 @@
...
+		poolsUseCase: puc,
+
+		bankQueryClient: banktypes.NewQueryClient(grpcClient),
+		stakingQueryClient: staking.NewQueryClient(grpcClient),
+		lockupQueryClient: lockup.NewQueryClient(grpcClient),
+		concentratedLiquidityQueryClient: concentratedLiquidity.NewQueryClient(grpcClient),
+	}, nil
+}
+
+func (p *passthroughUseCase) GetAccountCoinsTotal(ctx context.Context, address string) (sdk.Coins, error) {
```
p0mvn: We will need to unit-test this via mocks prior to merge

Could you take a look at how this is done in [this PR](https://github.com/osmosis-labs/sqs/pull/310)? See mocks in the `mocks` package and the way those are used in tests

ref: https://github.com/osmosis-labs/sqs/blob/ea77ffb9fde13736bb9c6f9fd202c5864f59df4f/domain/mocks/pool_liquidity_pricing_update_listenet_mock.go#L9

jonator:
Going to pause work on this as it is only needed to unblock the new portfolio page 

Tracking this work here:
https://linear.app/osmosis/issue/FE-539/add-total-coins-endpoint-to-sqs

jonator:
TODO

- [x] Needs unit tests
- [ ] Need to figure out if this is all coins in account
- [x] Consider breaking down by staking, pooled, unstaking, available, etc
- [x] Consider including price of each token with a given base denom (default USDC)

passthrough/usecase/passthrough_usecase.go
```diff
@@ -0,0 +1,140 @@
...
+
+			exitCoins, err := p.poolsUseCase.CalcExitCFMMPool(uint64(poolIDInt), balance.Amount)
+			if err != nil {
+				return nil, err
+			}
+			coins = coins.Add(exitCoins...)
+		} else {
+			coins = coins.Add(balance)
+		}
+	}
```
ValarDragon: This is whats going to be slow in the current edge function. (And the current costly thing in frontend, especially for accounts with slippage)

IMO should just ship without tallying balancer shares immediately, and let this come over time. Have this be in a second endpoint perhaps that can come async

passthrough/usecase/passthrough_usecase.go
ValarDragon: This should be split up into multiple different functions, one for each diff functionality, and get unit tested separately

passthrough/usecase/passthrough_usecase.go
ValarDragon: (If we wanted to ship portfolio w/ edge function only first, until this comes in the backend with proper testing)

passthrough/usecase/passthrough_usecase.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Staking
+	delegatedRes, err := p.stakingQueryClient.DelegatorDelegations(ctx, &staking.QueryDelegatorDelegationsRequest{
+		DelegatorAddr: address,
+	})
+	if err != nil {
+		return nil, err
+	}
+	undelegationRes, err := p.stakingQueryClient.DelegatorUnbondingDelegations(ctx, &staking.QueryDelegatorUnbondingDelegationsRequest{
+		DelegatorAddr: address,
```
ValarDragon: We should have a second pass later on, re-architecting these queries to be grpc-batched

ValarDragon: (Should also make queries in upstream SDK for just getting total tallies here)

p0mvn:
TODO:
- [x] docs
- [x] integration test
- [ ] observability
- [ ] Need to figure out if this is all coins in account

passthrough/delivery/http/passthrough_handler.go
```diff
@@ -0,0 +1,57 @@
...
+	return resourcePrefix + resource
+}
+
+// NewPassthroughHandler will initialize the pools/ resources endpoint
+func NewPassthroughHandler(e *echo.Echo, ptu mvc.PassthroughUsecase) {
+	handler := &PassthroughHandler{
+		PUsecase: ptu,
+	}
+
+	e.GET(formatPoolsResource("/portfolio-assets/:address"), handler.GetPortfolioAssetsByAddress)
```
p0mvn, line 32: @mattupham

domain/passthrough/passthrough_domain.go
```diff
@@ -0,0 +1,22 @@
...
+	TotalValueCap osmomath.Dec `json:"total_value_cap"`
+	// AccountCoinsResult represents coins only from user balances (contrary to TotalValueCap).
+	AccountCoinsResult []AccountCoinsResult `json:"account_coins_result"`
+}
+
+// AccountCoinsResult represents the coin balance as well as its capitalization value.
+type AccountCoinsResult struct {
+	Coin                sdk.Coin     `json:"coin"`
+	CapitalizationValue osmomath.Dec `json:"cap_value"`
+}
```
p0mvn, line 22: @mattupham API result format

PR #322, opened 2024-06-14
Orderbook support

iboss-ptk:


domain/errors.go
```diff
@@ -147,12 +150,33 @@ func (e ConcentratedTickModelNotSetError) Error() string {
...
 
-type AlloyTransmuterDataMissingError struct {
-	PoolId uint64
+// CosmWasmPoolType represents the type of a CosmWasm pool.
+type CosmWasmPoolType int
+
+const (
+	CosmWasmPoolAlloyTransmuter CosmWasmPoolType = iota
+	CosmWasmPoolOrderbook
+)
```
p0mvn: Should we encapsulate all other types under this?
- Transmuter v1
- Generalized

router/usecase/pools/pool_factory.go
```diff
@@ -125,6 +127,25 @@ func newRoutableCosmWasmPool(pool sqsdomain.PoolI, cosmWasmConfig domain.CosmWas
...
+
+			return &routableOrderbookPoolImpl{
+				ChainPool:     cosmwasmPool,
+				Balances:      balances,
+				TokenOutDenom: tokenOutDenom,
+				TakerFee:      takerFee,
+				SpreadFactor:  spreadFactor,
+				OrderbookData: model.Data.Orderbook,
+			}, nil
+		}
```
p0mvn: This method starts growing in complexity. Could you cover this new logic with a table-driven unit test?

If tests for other edge cases do not exist, please create a task to come back to this 🙏

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,223 @@
...
+
+// GetCodeID implements domain.RoutablePool.
+func (r *routableOrderbookPoolImpl) GetCodeID() uint64 {
+	return r.ChainPool.CodeId
+}
+
+// Determines order direction for the current orderbook given token in and out denoms
+// Returns:
+// - 1 if the order is a bid (buying token out)
+// - -1 if the order is an ask (selling token out)
```
p0mvn:
```suggestion
// - sqsdomain.BID (1) if the order is a bid (buying token out)
// - sqsdomain.ASK (-1) if the order is an ask (selling token out)
```

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,223 @@
...
+func amountToValue(amount osmomath.BigDec, price osmomath.BigDec, direction sqsdomain.OrderbookDirection) osmomath.BigDec {
+	switch direction {
+	case sqsdomain.ASK:
+		return amount.MulMut(price)
+	case sqsdomain.BID:
+		return amount.QuoMut(price)
+	default:
+		return osmomath.ZeroBigDec()
+	}
+}
```
p0mvn: Could we cover this method with the table-driven unit tests?

Please cover the invariants around parameter mutations.

For example cover with a test that `amount` is mutated

sqsdomain/cw_pool.go
```diff
@@ -62,6 +66,18 @@ func (model *CosmWasmPoolModel) IsAlloyTransmuter() bool {
 	return model.ContractInfo.Matches(name, constraints)
 }
 
+func (model *CosmWasmPoolModel) IsOrderbook() bool {
+	name := "crates.io:sumtree-orderbook"
+	version := ">= 0.1.0"
```
p0mvn: nit: define as constants at the top of the file

sqsdomain/cw_pool.go
```diff
@@ -80,3 +96,94 @@ type TransmuterAssetConfig struct {
 	// [more info](https://github.com/osmosis-labs/transmuter/tree/v3.0.0?tab=readme-ov-file#normalization-factors)
 	NormalizationFactor osmomath.Int `json:"normalization_factor"`
 }
+
+type OrderbookDirection int
+
+const (
+	BID OrderbookDirection = 1
+	ASK OrderbookDirection = -1
+)
```
p0mvn: We should put items into `sqsdomain` package only if they are used by both `osmosis` and `sqs` projects.

Otherwise, we should keep them under `domain`

Question - do we need all of this to be here or can we move some items to `domain`?

Feel free to close if need everything here 

The reason is that we want to keep `sqsdomain` as lean as possible so that we don't have to do `go.mod` updates frequently.

sqsdomain/cw_pool.go
```diff
@@ -80,3 +96,94 @@ type TransmuterAssetConfig struct {
...
+	return -1
+}
+
+type TickValues struct {
+	// Total Amount of Liquidity at tick (TAL)
+	// - Every limit order placement increments this value.
+	// - Every swap at this tick decrements this value.
+	// - Every cancellation decrements this value.
+	TotalAmountOfLiquidity osmomath.BigDec `json:"total_amount_of_liquidity"`
+}
```
p0mvn: Why does this have to be in a struct and not as `BigDec` directly?

Having this in a struct introduces light serialization overhead

p0mvn (commented):
Thanks for opening the PR! Nice job

I did a high-level skim.

I will do an in-depth review once the tests are added. Let's aim to maximize test coverage

sqsdomain/cw_pool.go
iboss-ptk: agreed, the original `TickValues` has a lot more field, this is all we need to sqs. Will remove the wrapped struct.

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: amount shouldn't be mutated, fixed that.

sqsdomain/cw_pool.go
iboss-ptk: will review the needs here.

router/usecase/pools/pool_factory.go
iboss-ptk: > If tests for other edge cases do not exist

would you mind elaborate on this?

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: apart from that I think this is well covered by test for spot price and calc function

router/usecase/pools/pool_factory.go
```diff
@@ -161,3 +145,66 @@ func newRoutableCosmWasmPool(pool sqsdomain.PoolI, cosmWasmConfig domain.CosmWas
...
+	}
+
+	return nil, nil
```
p0mvn: Could we return an error in this case? 


Returning nil, nil is an anti-pattern that encourages coupling this function to the context of its caller.

router/usecase/pools/pool_factory.go
```diff
@@ -101,31 +101,15 @@ func newRoutableCosmWasmPool(pool sqsdomain.PoolI, cosmWasmConfig domain.CosmWas
...
-				SpreadFactor:        spreadFactor,
-			}, nil
-		}
+	routablePool, err := newRoutableCosmWasmPoolWithCustomModel(pool, cosmwasmPool, cosmWasmConfig, tokenOutDenom, takerFee)
+	if err != nil {
+		return nil, err
 	}
+	if routablePool != nil {
+		return routablePool, nil
+	}
```
p0mvn: Related to my other comment - let's move this down to the bottom of `newRoutableCosmWasmPool` and make `newRoutableCosmWasmPoolWithCustomModel` return an error for unsupported pools.

Then, we can avoid the `nil, nil` situation and the check for `routablePool != nil`

router/usecase/pools/pool_factory.go
```diff
@@ -161,3 +145,66 @@ func newRoutableCosmWasmPool(pool sqsdomain.PoolI, cosmWasmConfig domain.CosmWas
 		PoolId:   cosmwasmPool.PoolId,
 	}
 }
+
+// newRoutableCosmWasmPoolWithCustomModel creates a new RoutablePool for CosmWasm pools that require a custom CosmWasmPoolModel.
+// errors if the pool matched criteria for a custom model, but the model does not have the required data.
+// returns a routable pool with a custom model if the pool matched criteria for a custom model and the model has the required data.
+// returns nil if the pool did not match criteria for a custom model.
+func newRoutableCosmWasmPoolWithCustomModel(
```
p0mvn, line 143: I also think this function would benefit from a unit test

router/usecase/pools/pool_factory.go
p0mvn: Let's create a test for `newRoutableCosmWasmPoolWithCustomModel` and a task to later unit test `newRoutableCosmWasmPool`

router/usecase/pools/routable_cw_orderbook_pool_test.go
```diff
@@ -0,0 +1,438 @@
...
+) sqsdomain.RoutablePool {
+	// TODO: replace this with orderbook, but this should work as mock for now
+	cosmwasmPool := s.PrepareCustomTransmuterPool(s.TestAccs[0], []string{tokenInDenom, tokenOutDenom})
+
+	poolType := cosmwasmPool.GetType()
+
+	mock := &mocks.MockRoutablePool{
+		ChainPoolModel: cosmwasmPool.AsSerializablePool(),
+		CosmWasmPoolModel: cosmwasmpool.NewCWPoolModel(
+			"crates.io:sumtree-orderbook", "0.1.0",
```
p0mvn: Let's please define this as a constant that we also use in the core implementation for semantic version checking

router/usecase/pools/routable_cw_orderbook_pool_test.go
```diff
@@ -0,0 +1,438 @@
...
+				},
+			},
+		),
+		PoolType: poolType,
+		TakerFee: takerFee,
+	}
+
+	routablePool, err := pools.NewRoutablePool(mock, tokenOutDenom, takerFee, domain.CosmWasmPoolRouterConfig{
+		OrderbookCodeIDs: map[uint64]struct{}{
+			defaultPoolID: {},
```
p0mvn:
```suggestion
			cosmwasmPool.GetId(): {},
```

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+	return r.SpreadFactor
+}
+
+// CalculateTokenOutByTokenIn implements sqsdomain.RoutablePool.
+// It calculates the amount of token out given the amount of token in for a concentrated liquidity pool.
+// Fails if:
+// - the underlying chain pool set on the routable pool is not of cosmwasm type
+// - fails to retrieve the tick model for the pool
+// - the provided denom pair is not supported by the orderbook
+// - runs out of ticks during swap (token in is too high for liquidity in the pool)
```
p0mvn: More edge cases are not captured by the spec. Please update 🙏

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+	return poolmanagertypes.CosmWasm
+}
+
+// GetSpreadFactor implements domain.RoutablePool.
+func (r *routableOrderbookPoolImpl) GetSpreadFactor() math.LegacyDec {
+	return r.SpreadFactor
+}
+
+// CalculateTokenOutByTokenIn implements sqsdomain.RoutablePool.
+// It calculates the amount of token out given the amount of token in for a concentrated liquidity pool.
```
p0mvn:
```suggestion
// It calculates the amount of token out given the amount of token in for an orderbook pool.
```

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+
+// Returns tick state index for the given ID
+func getTickIndexById(d *cosmwasmpool.OrderbookData, tickId int64) int {
+	for i, tick := range d.Ticks {
+		if tick.TickId == tickId {
+			return i
+		}
+	}
+	return -1
+}
```
p0mvn: We should maintain tick indexes for both bids and asks so that we don't have to do this search for performance reasons.

Alternatively, we could do binary search but I really think maintaining indexes is a more performant longer-term approach that we should strive towards right away

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+func (r *routableOrderbookPoolImpl) GetCodeID() uint64 {
+	return r.ChainPool.CodeId
+}
+
+// Determines order direction for the current orderbook given token in and out denoms
+// Returns:
+// - cosmwasmpool.BID (1) if the order is a bid (buying token out)
+// - cosmwasmpool.ASK (-1) if the order is an ask (selling token out)
+// - 0 if the order is not valid
+func (r *routableOrderbookPoolImpl) GetDirection(tokenInDenom, tokenOutDenom string) (domain.OrderbookDirection, error) {
```
p0mvn: Let's cover with unit tests please to max coverage

router/usecase/pools/routable_cw_orderbook_pool_test.go
```diff
@@ -0,0 +1,438 @@
...
+	}, domain.UnsetScalingFactorGetterCb)
+	s.Require().NoError(err)
+
+	return routablePool
+}
+
+func (s *RoutablePoolTestSuite) TestCalculateTokenOutByTokenIn_Orderbook() {
+	tests := map[string]struct {
+		tokenIn     sdk.Coin
+		tokenOut    sdk.Coin
```
p0mvn: Let's name this `expectedTokenOut` and put at the bottom of the test case definition for readability and style consistency (please see other test cases across the repository).

router/usecase/pools/routable_cw_orderbook_pool_test.go
```diff
@@ -0,0 +1,438 @@
...
+			tokenOut:    sdk.NewCoin(INVALID_DENOM, osmomath.NewInt(125)),
+			nextBidTick: 0,
+			nextAskTick: 0,
+			ticks:       []cosmwasmpool.OrderbookTick{},
+			expectError: domain.OrderbookPoolMismatchError{
+				PoolId:        defaultPoolID,
+				TokenInDenom:  BASE_DENOM,
+				TokenOutDenom: INVALID_DENOM,
+			},
+		},
```
p0mvn: Do you mind adding tests that exhaust liquidity in both directions?

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+
+		// Increment or decrement the current tick index depending on out order direction
+		switch directionOut {
+		case domain.BID:
+			tickIdx--
+		case domain.ASK:
+			tickIdx++
+		default:
+			return sdk.Coin{}, domain.OrderbookPoolInvalidDirectionError{Direction: directionIn}
+		}
```
p0mvn: nit: I think a simple if/else check would reduce cognitive overhead here

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return osmomath.BigDec{}, err
+	}
+
+	tickIdx, err := r.GetStartTickIndex(direction.Opposite())
+	if err != nil {
+		return osmomath.BigDec{}, err
+	}
+
+	tick := r.OrderbookData.Ticks[tickIdx]
```
p0mvn, line 169: This can be a separate function that is shared with ` CalculateTokenOutByTokenIn`

p0mvn (commented):
Nice work!

Mainly requesting more tests + tracking tick indexes rather than ticks so that we don't have to do iteration over ticks to search for the index.

Additionally, I think that we will need to come back to this after the mainntet cw pool is instantiated so that we can create s mock using `make sqs-update-mainnet-state` and add more higher level tests.

Additionally, we would need to come back to adding load tests specifically for orderbook pools/pairs.

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: I see this the opposite way, but this is minor, no strong opinion. will follow you here.

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: Let's not over abstract it here because both `directionIn` and `directionOut` are still being used down the line. We can have multiple return and such, but I think that's an overkill.

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: We don't even need whole idx -> id indexing since the use for this function is for finding nextBid / nextAsk index in ticks. We can just send nextBid / nextAsk index from ingester instead and perform (binary?) search only once over there.

router/usecase/pools/routable_cw_orderbook_pool.go
p0mvn: OK

domain/pools.go
```diff
@@ -56,3 +58,48 @@ var UnsetScalingFactorGetterCb ScalingFactorGetterCb = func(denom string) (osmom
...
+func (d *OrderbookDirection) String() string {
+	switch *d {
+	case BID:
+		return "BID"
+	case ASK:
+		return "ASK"
+	default:
+		return "UNKNOWN"
+	}
+}
```
p0mvn: Since there are only 2 possible values, let's make them boolean. That way, we don't need to handle this "UNKNOWN" case and generally improve readability. Let me know what you think

router/usecase/pools/pool_factory_test.go
```diff
@@ -0,0 +1,248 @@
...
+	panic("unimplemented")
+}
+
+func (m mockPool) SetTickModel(tickModel *sqsdomain.TickModel) error {
+	return nil
+}
+
+func (m mockPool) Validate(minUOSMOTVL osmomath.Int) error {
+	panic("unimplemented")
+}
```
p0mvn: AFAIR, there is already a mock in the `domain/mocks` package that we could reuse. However, if not, let's move this implementation there so that it can later be reused in other tests.

In the future, we can disable mocks from regular binary build with build tags if that's a concern. But it is fine for now IMO.

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,246 @@
...
+// Converts an amount of token in to the value of token out given a price and direction
+func convertValue(amount osmomath.BigDec, price osmomath.BigDec, direction domain.OrderbookDirection) osmomath.BigDec {
+	switch direction {
+	case domain.ASK:
+		return amount.Mul(price)
+	case domain.BID:
+		return amount.Quo(price)
+	default:
+		return osmomath.ZeroBigDec()
+	}
```
p0mvn, line 225: Having BID and ASK be booleans would remove the need for handling this default case as well as many other helper default cases down below.

router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -0,0 +1,246 @@
...
+
+	amountOutTotal := osmomath.ZeroBigDec()
+	amountInRemaining := osmomath.BigDecFromSDKInt(tokenIn.Amount)
+
+	// ASSUMPTION: Ticks are ordered
+	for amountInRemaining.GT(zeroBigDec) {
+		// Order has run out of ticks to iterate
+		if tickIdx >= len(r.OrderbookData.Ticks) || tickIdx < 0 {
+			return sdk.Coin{}, domain.OrderbookNotEnoughLiquidityToCompleteSwapError{PoolId: r.GetId(), AmountIn: tokenIn}
+		}
```
p0mvn, line 94: I'm worried about performance. Imagine that many users keep playing with the router, trying to estimate a large amount in that is guaranteed to exceed available liquidity in the direction of the swap.

With the current design, every time users type something different, we would iterate over all available ticks before reaching the end and concluding "no available liquidity".

Is it possible to store an accumulator that aggregates available liquidity in each direction? That way, we could use it to understand if the given token in amount exceeds the available liquidity immediately without having to iterate until the end.

Please let me know what you think.

p0mvn (commented):
Great work! Thank you for addressing the previous comments and adding more tests.

The only comment that is preventing me from hitting approve is the one related to performance. Let's please converge on this, and then we can push this forward.

I would appreciate if you could evaluate/address other remaining comments as well.

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: I think we can precompute amount in required to consume liquidity on all of the ticks and we can check on that. That has to be done from the ingester side then.

iboss-ptk: With that we can really index each tick by it's accumulated required amount in and binary search the tick to compute the last tick filled and avoid O(n) from router all together. Not sure if that's an overoptimization or not though. But I think it's possible.

p0mvn:
@iboss-ptk we also need to make sure that we appropriately ranked orderbook relative to other pools.

See this function: https://github.com/osmosis-labs/sqs/blob/ccb86f1f8b020ffba21c8b21f0b460ea4497093f/router/usecase/router.go#L112

Otherwise, orderbooks might be deprioritized relative to other pools

iboss-ptk:
> @iboss-ptk we also need to make sure that we appropriately ranked orderbook relative to other pools.
> 
> See this function:
> 
> https://github.com/osmosis-labs/sqs/blob/ccb86f1f8b020ffba21c8b21f0b460ea4497093f/router/usecase/router.go#L112
> 
> Otherwise, orderbooks might be deprioritized relative to other pools

are we doing the same thing as transmuter?

p0mvn:
> > @iboss-ptk we also need to make sure that we appropriately ranked orderbook relative to other pools.
> > See this function:
> > https://github.com/osmosis-labs/sqs/blob/ccb86f1f8b020ffba21c8b21f0b460ea4497093f/router/usecase/router.go#L112
> > 
> > Otherwise, orderbooks might be deprioritized relative to other pools
> 
> are we doing the same thing as transmuter?

I would probably rank orderbook a bit lower than CL to start

p0mvn (approved):
@iboss-ptk thank you for addressing comments.

I propose merging the current work and immediately following-up with the performance changes as follow-up PRs.

The reason is that we want to unblock concurrent integrations and testing via FE.

SQS risks becoming a bottleneck and the performance issues are not blocking the integrations.

Prior to merging this work, please add a documentation section summarizing the most important details of the orderbook integration in [this architecture document](https://github.com/osmosis-labs/sqs/blob/v25.x/docs/architecture/COSMWASM_POOLS.MD)

router/usecase/pools/routable_cw_orderbook_pool.go
p0mvn: @iboss-ptk I agree that the second approach is more performant but unlikely to be launch blocking. 

That being said, I think it would be beneficial to have this algorithm created so that we can minimize quote tail latencies. Performance of the router is critical.

Estimates over CL pools had been observed to be the bottleneck due to similar iterations, and I think efficiently designing orderbook will pay off down the road while also giving the flexibility to transfer this architecture to CL.

Let's break this down into 2 stages:
- Stage 1: only preventing not enough liquidity cases from iterating through all the ticks
- Stage 2: the aggregation and binary search mechanism you proposed

Please let me know if that sounds good.

Issue #279, opened 2024-06-03
[Q3] Auto-update asset list at regular intervals


p0mvn:
Currently, we need to redeploy SQS nodes to get SQS updates.

## DoD

* Mechanism to update asset list without redeployment
* Fault tolerance around new list being broken and falling back
* Observability
* Unit tests

p0mvn:
More context:

Let's have a config to enable this feature so that we can turn it on if something is breaking.

I suggest having a config value under [grpc-ingester](https://github.com/osmosis-labs/sqs/blob/3ddc4a6f46c4c8b59919d34158b5208253633aef/config.json#L49) named `asset-list-update-height-interval`.

If the value is non-zero, it would represent the frequency at which we refetch the asset list in terms of heights. For example, if interval is 1, means every height. If 5, every 5 height - 0, 5, 10 and so on.

It should be as simple as adding a call to a separate goroutine to refetch and update the asset list somewhere [here](https://github.com/osmosis-labs/sqs/blob/3ddc4a6f46c4c8b59919d34158b5208253633aef/ingest/usecase/ingest_usecase.go#L101) based on config.

deividaspetraitis:
Hey @p0mvn I am happy to pick this up. Thanks for providing some context and guidance!

deividaspetraitis:
This feature was delivered under [PR-375](https://github.com/osmosis-labs/sqs/pull/375), marking this task as completed.

PR #323, opened 2024-06-14
[Q3] Auto-update asset list at regular intervals

deividaspetraitis:
This PR implements updating asset list at regular intervals based on the procesing block in the ingester, please see #279 for more information.

## Implementation details

- As a side effect several more mocks were created in order to be able to test ingester use case, and existing mock was modified to support future testing requirements.

## Summary by CodeRabbit

- **New Features**
  - Introduced auto-updating asset list at regular intervals to enhance system functionality and data accuracy.
  - Added a new configuration parameter to define the block interval at which assets are updated.

- **Documentation**
  - Updated README with a corrected description for a token metadata endpoint and information on the new block interval configuration parameter.

- **Bug Fixes**
  - Corrected naming inconsistencies and reorganized parameter calls in various functions for clarity and improved functionality.

- **Tests**
  - Added new test functions to ensure assets are updated at specified intervals and to validate the new functionalities introduced.


tokens/usecase/tokens_usecase.go
```diff
@@ -357,6 +363,64 @@ func (t *tokensUseCase) getChainScalingFactorMut(precision int) (osmomath.Dec, b
 	return result, ok
 }
 
+var (
+	jobIsRunning   bool
+	jobIsrunningMu sync.Mutex
+)
+
+// StartAsRoutineIfNotRunning starts the job in a new goroutine if it is not already running.
+func StartAsRoutineIfNotRunning(job func()) bool {
```
deividaspetraitis: Those functions are public, because tests resides in other package, making original package to leak implementation details. Should I just create different file that would be within same package for example `tokens_usecase_i_test.go` where I could place tests?

tokens/usecase/tokens_usecase_test.go
```diff
@@ -343,3 +345,69 @@ func (s *TokensUseCaseTestSuite) TestGetPrices_Chain_PricingOptions() {
 		})
 	}
 }
+
+func TestDoEveryTick(t *testing.T) {
+	every, gen, iterations := rand.Intn(1000), rand.Intn(8932122), rand.Intn(1000)
+
+	var got int
+	for i := gen; i < gen+iterations+1; i++ {
+		tokensusecase.ExecuteOnTickInterval(uint64(i), every, func() {
```
deividaspetraitis: See https://github.com/osmosis-labs/sqs/pull/323/files#r1640012014

deividaspetraitis:
hey @p0mvn, while there are still several things left to implement in this PR, can I ask you for your early feedback at this point to make sure I am on the right track and consistent with the project code base ( I have left a few clarifying questions above )?

tokens/usecase/tokens_usecase.go
```diff
@@ -357,6 +363,64 @@ func (t *tokensUseCase) getChainScalingFactorMut(precision int) (osmomath.Dec, b
...
+
+// ExecuteOnTickInterval executes a callback at specified intervals.
+func ExecuteOnTickInterval(tick uint64, interval int, callback func()) {
+	if tick%uint64(interval) == 0 {
+		callback()
+	}
+}
+
+// UpdateAssetsAtHeightInterval updates assets at specified height intervals by running fn.
+// Function fn will be launched as separate goroutine.
```
deividaspetraitis: TODO: add docs; if `fn` ( `LoadTokensFromChainRegistryFunc` ) is already executing it will not be called twice.

tokens/usecase/tokens_usecase.go
```diff
@@ -110,18 +122,12 @@ func NewTokensUsecase(tokenMetadataByChainDenom map[string]domain.Token) mvc.Tok
...
-		poolDenomMetaData: sync.Map{},
+	t.tokenMetadataByChainDenom = tokenMetadataByChainDenom
+	t.humanToChainDenomMap = humanToChainDenomMap
+	t.precisionScalingFactorMap = precisionScalingFactors
 
-		chainDenoms:  chainDenoms,
-		coingeckoIds: coingeckoIds,
-	}
+	t.chainDenoms = chainDenoms
+	t.coingeckoIds = coingeckoIds
```
p0mvn: I think we should make all of these `sync.Map`. Then, iterate over reach updated map and update it element-by-element.

In the current approach, we would get a concurrent map read and write.

In the alternative approach of surrounding these with locks, I'm afraid that we would get a tail latency increase due to copying entire maps under locks.

tokens/usecase/tokens_usecase.go
```diff
@@ -357,6 +363,64 @@ func (t *tokensUseCase) getChainScalingFactorMut(precision int) (osmomath.Dec, b
...
+	}
+	jobIsRunning = true
+	jobIsrunningMu.Unlock()
+
+	go func() {
+		jobIsrunningMu.Lock()
+		job()
+		jobIsRunning = false
+		jobIsrunningMu.Unlock()
+	}()
```
p0mvn: Note that we don't really need this synchronization. We should be able to simply spin up a goroutine without any locks

tokens/usecase/tokens_usecase.go
```diff
@@ -357,6 +363,64 @@ func (t *tokensUseCase) getChainScalingFactorMut(precision int) (osmomath.Dec, b
...
+
+// UpdateAssetsAtHeightInterval updates assets at specified height intervals by running fn.
+// Function fn will be launched as separate goroutine.
+func UpdateAssetsAtHeightInterval(height uint64, n int, fn LoadTokensFromChainRegistryFunc, us mvc.TokensUsecase, chainRegistryAssetsFileURL string) {
+	ExecuteOnTickInterval(height, n, func() {
+		StartAsRoutineIfNotRunning(func() {
+			fn(us, chainRegistryAssetsFileURL) // nolint
+		})
+	})
+}
```
p0mvn: Consider simplifying for readability.

Suggest doing something as simple as:
```go
if tick%uint64(interval) == 0 {
  go func() {
      // Handle asset list update
  }()
}
```

app/sidecar_query_server.go
```diff
@@ -155,7 +155,27 @@ func NewSideCarQueryServer(appCodec codec.Codec, config domain.Config, logger lo
...
+					tokensUsecase.UpdateAssetsAtHeightInterval(
+						height,
+						200, // TODO config
+						tokensUsecase.LoadTokensFromChainRegistry,
+						tokensUseCase,
+						config.ChainRegistryAssetsFileURL,
+					)
+					logger.Info("callback block processing", zap.Uint64("height", height))
+				},
+			},
```
p0mvn: Not super sure why we need this wiring.

Can we instead have ingest handler call this:
```
tokensUsecase.UpdateAssetsAtHeightInterval(height)
```

For all the config parameters, we can set them on tokens use case in constructor

p0mvn:
A nit about mocks:

Let's say we have an `mvc.TokensUsecase` with 10 methods and `mvc.IngestUsecase` depends on the `mvc.TokensUsecase` but used only 2 methods.

Then, instead of creating a mock with 10 methods, we can make `mvc.IngestUsecase` depend on a separate interface with only 2 methods that `mvc.TokensUsecase` also implements.

This would allow us to customize the mock for this specific need without too much boilerplate.

Please let me know if anything doesn't make sense

cryptomatictrader:
@deividaspetraitis , I have a question about the implementation, does it update the entire asset list on every frequency, regardless whether there is any change? I wonder if it's worth adding the logic of storing the file size / checksum of the file in memory so that it won't even bother to touch the asset list at all?

tokens/usecase/tokens_usecase.go
deividaspetraitis: Implementation reflects the idea that only one assets update go routine can be running at any given time. For example if routine is already running second routine would be not launched to avoid DOS'ing external service.

I thought it can be useful for some edge cases:

- Wrong configuration of the update frequency ( for example every height )
- Network latency, some routine may be still in progress while other is started
- TCP/IP connections ( sockets ) are files, unlimited number of open descriptor might lead to problems

Please let me know would you would like still to proceed to remove it and I will address that.

deividaspetraitis:
> A nit about mocks:
> 
> Let's say we have an `mvc.TokensUsecase` with 10 methods and `mvc.IngestUsecase` depends on the `mvc.TokensUsecase` but used only 2 methods.
> 
> Then, instead of creating a mock with 10 methods, we can make `mvc.IngestUsecase` depend on a separate interface with only 2 methods that `mvc.TokensUsecase` also implements.
> 
> This would allow us to customize the mock for this specific need without too much boilerplate.
> 
> Please let me know if anything doesn't make sense

Just to confirm I get your idea right, continuing building on your idea, instead of:

```
func NewTokensUsecase(mvc.IngestUsecase) ... // mvc.IngestUsecase is 10 methods inferface, of which only 2 is used in constructed `TokensUsecase`
```

Have:

```
func NewTokensUsecase(mvc.Ingest2Methods) ... // mvc.Ingest2Methods is 2 methods inferface, of which all 2 are utilised in constructed `TokensUsecase`
```

Is that right?

deividaspetraitis:
> @deividaspetraitis , I have a question about the implementation, does it update the entire asset list on every frequency, regardless whether there is any change? I wonder if it's worth adding the logic of storing the file size / checksum of the file in memory so that it won't even bother to touch the asset list at all?

@cryptomatictrader, thank you for your time reviewing this PR and for feedback! You're totally right, current approach does the update regardless whether there is any change. Additionally, I think, we could try to determine whether we want to download assets data at all, there must be some HTTP header indicating whether contents has changed or not? In any case, it's probably question of priorities and I am happy to improve current implementation if you guys think it's worth spending the time on that at this point.

cryptomatictrader:
> > @deividaspetraitis , I have a question about the implementation, does it update the entire asset list on every frequency, regardless whether there is any change? I wonder if it's worth adding the logic of storing the file size / checksum of the file in memory so that it won't even bother to touch the asset list at all?
> 
> @cryptomatictrader, thank you for your time reviewing this PR and for feedback! You're totally right, current approach does the update regardless whether there is any change. Additionally, I think, we could try to determine whether we want to download assets data at all, there must be some HTTP header indicating whether contents has changed or not? In any case, it's probably question of priorities and I am happy to improve current implementation if you guys think it's worth spending the time on that at this point.

Thanks for the suggestion. The asset list is fetched from github site and I doubt they have some built-in header for us to perform some pre-flight check, plus the overhead of loading such file is minimal (just checked and it's around 150ms total response time). 

I think it's worth optimizing the update logic such that if there is no content change at all, we just skip the entire update process. We can do this by checking the total byte size of the response. What do you think?

deividaspetraitis:
> > > @deividaspetraitis , I have a question about the implementation, does it update the entire asset list on every frequency, regardless whether there is any change? I wonder if it's worth adding the logic of storing the file size / checksum of the file in memory so that it won't even bother to touch the asset list at all?
> 
> > 
> 
> > @cryptomatictrader, thank you for your time reviewing this PR and for feedback! You're totally right, current approach does the update regardless whether there is any change. Additionally, I think, we could try to determine whether we want to download assets data at all, there must be some HTTP header indicating whether contents has changed or not? In any case, it's probably question of priorities and I am happy to improve current implementation if you guys think it's worth spending the time on that at this point.
> 
> 
> 
> Thanks for the suggestion. The asset list is fetched from github site and I doubt they have some built-in header for us to perform some pre-flight check, plus the overhead of loading such file is minimal (just checked and it's around 150ms total response time). 
> 
> 
> 
> I think it's worth optimizing the update logic such that if there is no content change at all, we just skip the entire update process. We can do this by checking the total byte size of the response. What do you think? 

That definitely sounds like a plan, thank you for sharing ideas and testing possible approaches. Let me implement your suggested check once I am back working on this task.

tokens/usecase/tokens_usecase.go
p0mvn: I see. I would like to recommend considering using a worker pool [design pattern](https://shopify.engineering/leveraging-go-worker-pools) to address these edge cases.

We have the barebones abstractions defined [here](https://github.com/osmosis-labs/sqs/blob/v25.x/domain/workerpool/worker_pool.go)

By leveraging the abstraction, we will reuse the existing components, reducing complexity and addressing the raised concerns by configuring a pool with 1 concurrent worker.

However, I think that we can derisk them with a wider height interval that we can control with tests and observability.

I'll leave the worker pool decision up to you but I think that we should favor code simplicity here generally and derisk any concerns with tests and observability.

tokens/usecase/tokens_usecase.go
deividaspetraitis: Removed this code, no longer relevant.

tokens/usecase/tokens_usecase.go
deividaspetraitis: For now I have removed to keep things simple as per you suggestion, we can always add this feature in the future if there is a need.

app/sidecar_query_server.go
deividaspetraitis: Original idea I had was to have more generic callback mechanism, where `ingester` just calls any number of `callback` functions with signature of `func(height uint64)`, and in turn specifics of constructing actual wiring were moved to the program entry point. 

As per your suggestion I have updated to just call `tokensUsecase.UpdateAssetsAtHeightInterval(height)` in the ingest handler.

deividaspetraitis: Original idea I had was to have more generic callback mechanism, where `ingester` just calls any number of `callback` functions with signature of `func(height uint64)`, and in turn specifics of constructing actual wiring were moved to the program entry point. 

As per your suggestion I have updated to just call `tokensUsecase.UpdateAssetsAtHeightInterval(height)` in the ingest handler.

tokens/usecase/tokens_usecase.go
deividaspetraitis: Thanks, I have updated code to reflect your idea. :)

tokens/usecase/tokens_usecase.go
```diff
@@ -349,26 +391,69 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
...
+	if !ok {
+		return osmomath.Dec{}, false
+	}
+
+	return v, true
+}
+
+// UpdateAssetsAtHeightInterval updates assets at configured height interval.
+// Internally, it calls LoadTokensFromChainRegistry as a goroutine.
+func (t *tokensUseCase) UpdateAssetsAtHeightInterval(height uint64) {
```
deividaspetraitis: Should we cover with tests this method? Any suggestions how we can do that?

tokens/usecase/tokens_usecase.go
```diff
@@ -349,26 +391,69 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
...
+	if t.lastLoadTokensFromChainRegistryHash != hash {
+		t.LoadTokens(tokens)
+	}
+
+	return nil
 }
 
 // GetTokensFromChainRegistry fetches the tokens from the chain registry.
 // It returns a map of tokens by chain denom.
-func GetTokensFromChainRegistry(chainRegistryAssetsFileURL string) (map[string]domain.Token, error) {
```
deividaspetraitis, line 358: Should we cover this function with tests?

deividaspetraitis:
Hey @p0mvn, @cryptomatictrader I have addressed your previous comments, can I ask please for one more round of reviews? I have also opened a few threads where your input would be appreciated. This PR is marked now as "ready for reviews", since based on your previous reviews I feel that we are getting close finishing it. :) Thanks!

tokens/usecase/tokens_usecase.go
p0mvn: nit: I think having this specified for the caller would be helpful
```suggestion
func (t *tokensUseCase) UpdateAssetsAtHeightIntervalAsync(height uint64) {
```

p0mvn: Yes. For the context of this method, we simply need to validate the following:
- `LoadTokensFromChainRegistry` is called at the expected height interval
- `LoadTokensFromChainRegistry` is not called at the expected height interval

Other invariants are to be validated by unit tests of the downstream methods.

We might have to rely on mocks to properly set this up.

tokens/usecase/tokens_usecase.go
p0mvn: I think the ROI of covering this method by tests would be low assuming that we diligently unit test other methods.

It is hard to set up the test for this since we would have to mock out the HTTP request stuff. 

We also have higher-level functional tests for this component ensuring that this parses as intended. So I would advise to not spend too much time at this time.

tokens/usecase/tokens_usecase.go
```diff
@@ -349,26 +391,69 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
...
+	if err != nil {
+		return err
+	}
+
+	if t.lastLoadTokensFromChainRegistryHash != hash {
+		t.LoadTokens(tokens)
+	}
+
+	return nil
 }
```
p0mvn: I think we should also unit-test this method. We should define `getTokensFromChainRegistry` as a field on `tokensUseCase`.

In core logic, we can wire up the actual implementation.

In the test logic, we can monkey-patch a mock.

Let me know if you'd like to have more pointers on how to get this working

tokens/usecase/tokens_usecase.go
```diff
@@ -411,8 +496,18 @@ func (t *tokensUseCase) RegisterPricingStrategy(source domain.PricingSourceType,
...
-	return ok && !metaData.IsUnlisted
+	metaData, ok := t.tokenMetadataByChainDenom.Load(chainDenom)
+	if !ok {
+		return false
+	}
+
+	v, ok := metaData.(domain.Token)
+	if !ok {
+		return false
+	}
```
p0mvn, line 446: Since this method starts growing in complexity, I suggest we add simple unit tests to ensure maximal test coverage while validating that the proper typing invariants are maintained.

tokens/usecase/tokens_usecase.go
```diff
@@ -446,9 +541,15 @@ func (t *tokensUseCase) IsValidPricingSource(pricingSource int) bool {
...
-	} else {
+	coingeckoId, found := t.coingeckoIds.Load(chainDenom)
+	if !found {
 		return "", fmt.Errorf("chain denom not found in chain registry")
 	}
+
+	v, ok := coingeckoId.(string)
+	if !ok {
+		return "", fmt.Errorf("coingecko id for chain denom (%v) is not of type string", chainDenom)
+	}
```
p0mvn, line 491: Ditto for covering with tests and all other similar instances.

ingest/usecase/ingest_usecase_test.go
```diff
@@ -366,6 +370,48 @@ func (s *IngestUseCaseTestSuite) TestTransferDenomLiquidityMap() {
...
+
+	want := rand.Intn(100)
+	for i := 0; i < want; i++ {
+		err = ingester.ProcessBlockData(context.TODO(), 1, nil, nil)
+		s.Require().NoError(err)
+	}
+
+	if want != got {
+		s.T().Fatalf("got %v, want %v", want, got)
+	}
```
p0mvn: I know this is probably code generated but we can do:
```go
s.Require().Equal(want, got)
```

p0mvn (commented):
Nice work!

Mostly had comments around adding tests but this is looking solid so far 👍

tokens/usecase/tokens_usecase.go
deividaspetraitis: I have covered this method with tests: https://github.com/osmosis-labs/sqs/pull/323/files#diff-393639508320686c9dbe3457d2678cac937a6294c260d8f30a59047d394f1193R548 

Note that I am unable cover with type assertion tests because  `tokensUseCase.coingeckoIds` is unexported and tests resides in different package meaning I am unable to load it with with invalid types ( for example `int` instead of `domain.Token` ). 

There must be some reasoning behind decision making tests to live in separate package, thus I just want to learn is current coverage enough or we should cover mentioned cases as well? 

IMHO is that tests living in the same packages would simplify testing in overall and allow to cover such edge cases.

tokens/usecase/token_registry_test.go
```diff
@@ -0,0 +1,61 @@
+package usecase
+
+import (
+	"testing"
+
+	"github.com/osmosis-labs/sqs/domain"
+)
+
+func TestFetchAndUpdateTokens(t *testing.T) {
```
deividaspetraitis, line 9: This is very related with this comment: https://github.com/osmosis-labs/sqs/pull/323/files#r1665915262

I was not able to find elegant apprach how to write tests against `ChainRegistryHTTPFetcher` to check its internal state `initialHash` value without making test to live in the same packgae as t `ChainRegistryHTTPFetcher` implementation.  One of the options could possibly be to add public method to return `initialHash` but that would leak implementation details, with larger data structures that would result in too verbose API.

I understand that this approach is changing course of current approach and I am more than happy to align to it if we can came up with some ideas.

tokens/usecase/tokens_usecase.go
deividaspetraitis: Thanks, I have added test coverage for this method!

deividaspetraitis: Thanks, it's updated including in all other relevant places.

tokens/usecase/tokens_usecase.go
deividaspetraitis: Please see my comment on implementation: https://github.com/osmosis-labs/sqs/pull/323#discussion_r1665920048

app/sidecar_query_server.go
```diff
@@ -90,13 +90,22 @@ func NewSideCarQueryServer(appCodec codec.Codec, config domain.Config, logger lo
...
+	// TODO: Make the max number of tokens configurable
+	tokensUseCase := tokensusecase.NewTokensUsecase(
+		tokenMetadataByChainDenom,
+		config.UpdateAssetsHeightInterval,
+		tokensusecase.NewChainRegistryHTTPFetcher(
+			config.ChainRegistryAssetsFileURL,
+			tokensusecase.GetTokensFromChainRegistry,
+		),
+		logger,
+	)
```
p0mvn: nit for readability
```go
	// Initialize chain registry HTTP fetcher
	chainRegistryHTTPFetcher := tokensusecase.NewChainRegistryHTTPFetcher(
		config.ChainRegistryAssetsFileURL,
		tokensusecase.GetTokensFromChainRegistry,
	)

	// Initialized tokens usecase
	tokensUseCase := tokensusecase.NewTokensUsecase(
		tokenMetadataByChainDenom,
		config.UpdateAssetsHeightInterval,
		chainRegistryHTTPFetcher,
		logger,
	)
```

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+func NewChainRegistryHTTPFetcher(registryURL string, getTokensFromChainRegistry GetTokensFromChainRegistryFunc) *ChainRegistryHTTPFetcher {
+	return &ChainRegistryHTTPFetcher{
+		getTokensFromChainRegistry: getTokensFromChainRegistry,
+		registryURL:                registryURL,
+	}
+}
+
+// FetchAndUpdateTokensFunc is a FetchAndUpdateTokens function signature.
+type FetchAndUpdateTokensFunc func(loadTokens func(tokens map[string]domain.Token)) error
+
```
p0mvn: afaict this is unused

```suggestion
```

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+// FetchAndUpdateTokens fetches tokens from the chain registry and loads by calling loadTokens  function.
+// In case there were no changes since last fetch, it does not call loadTokens.
+func (f *ChainRegistryHTTPFetcher) FetchAndUpdateTokens(loadTokens LoadTokensFunc) error {
```
p0mvn, line 81: Could we instead propagate this in the constructor of `ChainRegistryHTTPFFetcher`?

ingest/usecase/ingest_usecase_test.go
```diff
@@ -366,6 +370,48 @@ func (s *IngestUseCaseTestSuite) TestTransferDenomLiquidityMap() {
...
+
+	want := rand.Intn(100)
+	for i := 0; i < want; i++ {
+		err = ingester.ProcessBlockData(context.TODO(), 1, nil, nil)
+		s.Require().NoError(err)
+	}
+
+	if want != got {
+		s.Require().Equal(want, got)
+	}
```
p0mvn:
```suggestion
	s.Require().Equal(want, got)
```

ingest/usecase/ingest_usecase_test.go
```diff
@@ -366,6 +370,48 @@ func (s *IngestUseCaseTestSuite) TestTransferDenomLiquidityMap() {
...
+			},
+		},
+		&mocks.CandidateRouteSearchDataWorkerMock{},
+		logger,
+	)
+	s.Require().NoError(err)
+
+	want := rand.Intn(100)
+	for i := 0; i < want; i++ {
+		err = ingester.ProcessBlockData(context.TODO(), 1, nil, nil)
```
p0mvn: Please consider creating variables for the inputs so that it is clearer what it is that we are propagating as inputs

tokens/usecase/tokens_usecase_test.go
```diff
@@ -539,5 +542,98 @@ func (s *TokensUseCaseTestSuite) TestGetMinPoolLiquidityCap() {
...
+
+// FetchAndUpdateTokens implements the tokensusecase.TokenLoader interface.
+func (m *mockTokenLoader) FetchAndUpdateTokens(loadTokens tokensusecase.LoadTokensFunc) error {
+	m.callback()
+	return nil
+}
+
+// TestUpdateAssetsAtHeightIntervalAsync tests the async update of assets at height interval.
+func (s *TokensUseCaseTestSuite) TestUpdateAssetsAtHeightIntervalAsync() {
+	logger, err := log.NewLogger(false, "", "")
```
p0mvn: There is a `log.NoOpLogger` that can be helpful in tests

ingest/usecase/ingest_usecase_test.go
```diff
@@ -366,6 +370,48 @@ func (s *IngestUseCaseTestSuite) TestTransferDenomLiquidityMap() {
...
+			UpdatePricesAsyncFunc: func(height uint64, uniqueBlockPoolMetaData domain.BlockPoolMetadata) {
+				// do nothing
+			},
+		},
+		&mocks.CandidateRouteSearchDataWorkerMock{},
+		logger,
+	)
+	s.Require().NoError(err)
+
+	want := rand.Intn(100)
```
p0mvn: Let's avoid pseudo-randomness in tests with a deterministic seed to prevent flakiness

tokens/usecase/tokens_usecase.go
p0mvn: It is possible to mock out this test case by creating a test-specific method in `export_test.go`

```go
func (t *tokensUseCase) PutInvalidTypeTokenMetadata(key string, value any) {
	t.tokenMetadataByChainDenom.Store(key, value)
}
```

Note, that since this method exists in file `..._test.go` but `usecase` package, it would only be compiled for tests but not into the prod binary.

tokens/usecase/tokens_usecase.go
```diff
@@ -349,40 +387,29 @@ func (t *tokensUseCase) getPricesForBaseDenom(ctx context.Context, baseDenom str
...
+	if height%uint64(t.updateAssetsHeightInterval) == 0 {
+		go func() {
+			if err := t.tokenLoader.FetchAndUpdateTokens(t.LoadTokens); err != nil {
+				t.logger.Error("error loading tokens from chain registry", zap.Error(err))
+			}
+		}()
 	}
-
-	return tokensByChainDenom, nil
 }
```
p0mvn: Consider simplifying to the following:
```go
// UpdateAssetsAtHeightIntervalAsync updates assets at configured height interval.
// Internally, it calls LoadTokensFromChainRegistry as a goroutine.
func (t *tokensUseCase) UpdateAssetsAtHeightIntervalAsync(height uint64) {
	if height%uint64(t.updateAssetsHeightInterval) == 0 {
		go func() {
			if err := t.updateAssetsAtHeightIntervalSync(height); err != nil {
				// TODO: add prometheus error counter so that we can create an alert for observability
				t.logger.Error("error loading tokens from chain registry", zap.Error(err))
			}
		}()
	}
}

func (t *tokensUseCase) updateAssetsAtHeightIntervalSync(height uint64) error {
	if height%uint64(t.updateAssetsHeightInterval) == 0 {
		if err := t.tokenLoader.FetchAndUpdateTokens(t.LoadTokens); err != nil {
			return err
		}
	}
	return nil
}
```

Note that by having a separate sync version and returning an error, we improve the testability of the method, allowing for the ability to set up table-driven test cases and validate all branches.

Also, please note my TODO about the Prometheus metric

tokens/usecase/tokens_usecase_test.go
```diff
@@ -539,5 +542,98 @@ func (s *TokensUseCaseTestSuite) TestGetMinPoolLiquidityCap() {
...
+
+	mock := &mockTokenLoader{
+		callback: callback,
+	}
+
+	tokensUsecase := tokensusecase.NewTokensUsecase(nil, updateAssetsHeightInterval, mock, logger)
+
+	for i := gen; i < gen+iterations+1; i++ {
+		tokensUsecase.UpdateAssetsAtHeightIntervalAsync(uint64(i))
+	}
```
p0mvn: Please see my other comment about separating a sync version of this so that we can set up a table-driven test case

tokens/usecase/tokens_usecase_test.go
```diff
@@ -539,5 +542,98 @@ func (s *TokensUseCaseTestSuite) TestGetMinPoolLiquidityCap() {
...
+	m.callback()
+	return nil
+}
+
+// TestUpdateAssetsAtHeightIntervalAsync tests the async update of assets at height interval.
+func (s *TokensUseCaseTestSuite) TestUpdateAssetsAtHeightIntervalAsync() {
+	logger, err := log.NewLogger(false, "", "")
+	s.Require().NoError(err)
+
+	updateAssetsHeightInterval, gen, iterations := rand.Intn(100), rand.Intn(8932122), rand.Intn(1000)
```
p0mvn: Ditto for using deterministic seeds as needed. Generally, I recommend opting for table-driven test cases , where possible, making the async version a simple propagator/wrapper to a heavily unit-tested synchronous method

tokens/usecase/tokens_usecase.go
p0mvn: Note that this ignores one of my previous comments about initializing `t.tokenLoader` with `t.LoadTokens` rather than propagating it for every call.

p0mvn (commented):
Good work!

Getting close. Once the comments from the current review round are addressed, I expect to leave another cycle of comments focusing more on tests. Please let me know once ready.

Also, if you find many comments overwhelming, please feel free to open a fresh PR to reduce the comment clutter (but up to you 👍 )

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/osmosis-labs/sqs/domain"
+)
+
+// GetTokensFromChainRegistryFunc is a GetTokensFromChainRegistry function signature.
+type GetTokensFromChainRegistryFunc func(chainRegistryAssetsFileURL string) (map[string]domain.Token, string, error)
+
+// GetTokensFromChainRegistry fetches the tokens from the chain registry.
+// It returns a map of tokens by chain denom.
+func GetTokensFromChainRegistry(chainRegistryAssetsFileURL string) (map[string]domain.Token, string, error) {
```
cryptomatictrader, line 18: Can we issue `response.Body.Close()` only once, and read the `response.Body` once ? Or there is a reason for that? I'm not sure what is the purpose of line 34 - line 39. Could you please explain? Thank you!

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	defer response.Body.Close()
+
+	// Decode the JSON data
+	var assetList AssetList
+	err = json.NewDecoder(response.Body).Decode(&assetList)
+	if err != nil {
+		return nil, "", err
+	}
+
```
cryptomatictrader, line 32: Start of the code needs explanation.

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+	data, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, "", err
+	}
+	response.Body.Close() //  close old response body
+	response.Body = io.NopCloser(bytes.NewBuffer(data))
+
```
cryptomatictrader, line 42: End of the code needs explanation

tokens/usecase/token_registry.go
```diff
@@ -0,0 +1,94 @@
...
+// FetchAndUpdateTokens fetches tokens from the chain registry and loads by calling loadTokens  function.
+// In case there were no changes since last fetch, it does not call loadTokens.
+func (f *ChainRegistryHTTPFetcher) FetchAndUpdateTokens(loadTokens LoadTokensFunc) error {
+	tokens, hash, err := f.getTokensFromChainRegistry(f.registryURL)
+	if err != nil {
+		return err
+	}
+
+	if f.lastFetchHash != hash {
```
cryptomatictrader, line 87: 👍 Nice optimization

cryptomatictrader (commented):
Nice work! Some minor issue / question to be addressed. Thank you!

tokens/usecase/tokens_usecase.go
deividaspetraitis: I just  tried  to go wtih your suggested approach and noticed that `tokensusecase.NewTokensUsecase` returns `mvc.TokensUsecase`, thus such method type is not available. If we want to go this route we can try return `TokensUseCase`  as contrete and exported type and then cast it in tests:

```
func NewTokensUsecase(tokenMetadataByChainDenom map[string]domain.Token, updateAssetsHeightInterval int, tokenRegistryLoader TokenRegistryLoader, logger log.Logger) mvc.TokensUsecase {
  ....
  return &TokensUseCase{...}
 ....
}
```

Tests:

```
usecase := tokensusecase.NewTokensUsecase(....)
usecase, ok := usecase.(tokensusecase.TokensUseCase)
```

Let me know in what appraoch you would favour.

tokens/usecase/token_registry.go
deividaspetraitis: That was initial idea, however that approach led to some challanges when I was trying to came up how I can test `LoadTokensFunc` and wire everything.

When we construct `ChainRegistryHTTPFFetcher` we would need to pass `tokensUsecase.LoadTokens` method, but when we are constructing we need to pass `tokensUsecase`:

```go
	tokensUseCase := tokensusecase.NewTokensUsecase(
		...
		chainRegistryHTTPFetcher,
		...
	)

	// Initialize chain registry HTTP fetcher
	chainRegistryHTTPFetcher := tokensusecase.NewChainRegistryHTTPFetcher(
		...
		tokensUseCase.LoadTokens
		...
	)
```

One possible approach could be to just add extra method on ` tokensusecase`, such as `tokensusecase.SetTokenRegistryLoader(TokenRegistryLoader)`, but I did not done that for several reasons:

1. Use case would more difficult to initialise ( more wiring )
2. Use case might not work as not expeted in case someone forgot to set TokenRegistryLoader
3. I would still need to initialise `tokenusecase` first to have a reference to `tokensUsecase.LoadTokens` method

Let me know if I should proceed this or any other way! :rocket:

tokens/usecase/token_registry.go
deividaspetraitis: Oh, that's a good catch, thank you! I am not sure how I have ended up with this piece of code. I have updated response to be read once and then data to be reused for several operations. Please let me know if that makes sense?

tokens/usecase/tokens_usecase.go
deividaspetraitis: Totally agree, that simplies a lot of the logic in tests as well. Thanks for suggestion, implemened and pushed relevant changes!

deividaspetraitis:
I am closing this pull request in favour of new new one, since this one is starting to get hard to follow.

PR #329, opened 2024-06-18
feat: Add fault tolerance in candidate route conversion when pools are breaking

p0mvn:
We had a production incident earlier where a migrated code ID and long candidate route caching led to this component bubbling up an error in a single route up top, completely failing client requests when alternatives were available.

Instead, we should be skipping the violating routes silently in such cases.

## Testing

Added a unit test


cryptomatictrader (approved):
LGTM. Thanks!

PR #333, opened 2024-06-19
Alessandro/add commit sha integration test action

alessandrolomanto:
This PR introduces the following changes:

- Update to `.github/workflows/integration-test.yml`:

  - Added a new optional input `commit_sha` that has a default value an is optional.

- Update to `.github/workflows/push-dev-docker-images.yml`:
  - Modified the line that generates the `temp.json` file to include the `commitsha` field with the value of the commit id.
  
  This PR is to run automatically the Integration Test Suite after stage deploy

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -25,6 +25,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.commit_sha || github.repository_default_branch }}
       
       - name: Set up Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


niccoloraspa:
We need to checkout that commit (if present) in:
https://github.com/osmosis-labs/sqs/blob/9a3048a0e31367dadecbe8954c0f7ffaf1360e7d/.github/workflows/integration-test.yml#L25-L28

niccoloraspa:
```yaml
steps:
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          ref: ${{ github.event.inputs.commit_sha || github.repository_default_branch }}
```

p0mvn (approved):

PR #335, opened 2024-06-20
Hotfix for Alloyed pool swaps failing due to not having LP share in balances

p0mvn:
The alloyed LP share is not present in the balances, causing a routing bug below.

Longer term we should solve this by ingesting the LP share balance for the alloyed pool but this is a quick hot fix for now.

image


iboss-ptk (approved):

cryptomatictrader (approved):
LGTM. Thank you!

PR #338, opened 2024-06-21
fix: remove token in check in candidate route search

p0mvn:
Discussed with @iboss-ptk 

This check is redundant and can be removed.


iboss-ptk (dismissed):

p0mvn:
I ended up adding alloyed lp share to the pool denoms but not the balances. Unsure if that's the right approach. I think we can close for now and revisit later

PR #332, opened 2024-06-19
feat: Pool liquidity capitalization/pricing

p0mvn:
https://linear.app/osmosis/issue/DATA-85/pool-liquidity-pricing

This PR implements pool liquidity capitalization/pricing.

As part of this effort, we now pre-compute USDC-denominated liquidity for a pool at the end of every block.

Suggest reviewing the `docs/architecture/ingest.md` to refresh understanding of how the worker pipelining works.

## Checklist
- [x] Fix test failures
   * RepriceDenomsMetadata - needs updates to using `UpdatedDenoms`
   * `repricePoolLiquidityCap`
- [x] Add logger to pool liquidity pricing worker
- [x] Unit test
   * CreatePoolDenomMetaData
- [x] Integration test
- [x] Make pool liquidity pricer worker concurrent again 

## Tests
- Unit tested with mocks
- Integration test added. Queries each pool and compares pool liquidity capitalization with Numia data.
- Tested on a node

PR #343, opened 2024-06-23
refactor/test/fix: liquidity pricer

p0mvn:
Refactor & progress towards pool liquidity capitalization/pricing

The core of the change focuses on creating better `PriceCoin` and `PriceBalances` APIs on the liquidity price and covering them with more tests.

Extracted from: https://github.com/osmosis-labs/sqs/pull/332 for ease of review.

As an outcome, identified and fixed an overapplying quote denom scaling factor bug in edge cases. It was unnoticed since this logic is not relied on in production yet.

Tested on the node

### DIFF
--- a/domain/pricing.go
+++ b/domain/pricing.go
@@ -149,11 +149,6 @@ type PricingUpdateListener interface {
 type PoolLiquidityPricerWorker interface {
 	// Implements PricingUpdateListener
 	PricingUpdateListener
-	// ComputeLiquidityCapitalization computes the capitalization of the liquidity for the given denom
-	// using the total liquidity and the price.
-	// Returs zero if the price is zero or if there is any internal error.
-	// Otherwise, returns the computed liquidity capitalization from total liquidity and price.
-	ComputeLiquidityCapitalization(denom string, totalLiquidity osmomath.Int, price osmomath.BigDec) osmomath.Int
 
 	// RepriceDenomMetadata reprices the token liquidity metadata for the denoms updated within the block.
 	// Returns the updated token metadata.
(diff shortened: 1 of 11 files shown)


domain/pricing.go
```diff
@@ -175,22 +170,27 @@ type PoolLiquidityPricerWorker interface {
 	RegisterListener(listener PoolLiquidityComputeListener)
 }
 
-// DenomPriceInfo defines the price information for the base denom.
 type DenomPriceInfo struct {
-	// Price is the price of the base denom.
-	Price osmomath.BigDec
-	// ScalingFactor is the scaling factor for the base denom.
+	Price         osmomath.BigDec
```
cryptomatictrader, line 174: Just curious - what is the general practice when using osmomath vs native (coz I think I will face some issue when doing integration test for dexscreener when we have data passing around node => pubsub => firestore) ? For price, can we not use osmomath.BigDec and just native data type?

cryptomatictrader: I see in most of our code, we tend to use osmomath by default vs native data type, and I think there must be some reasons for that (added convenient methods?)

cryptomatictrader (approved):
LGTM. Thank you!

domain/pricing.go
p0mvn: There are 2 reasons:
- We need cross-platform determinism that native floats do not provide. As a result, we use fixed point precision decimal types
- Size of the data. Prices can be large enough to exceed the max float value

Here in SQS, since we depend on the chain APIs, we simply inherit all of the types for simplicity.

PR #355, opened 2024-06-27
docs: candidate route optimization & demo

p0mvn:
Tracks the following milestone under performance improvements: https://linear.app/osmosis/issue/DATA-174/[candidaterouteopt]-optimize-candidate-routes

The core of the implementation is to be accomplished in follow-up PRs.

Review suggestion:
- `ingest.md` diff
- `router.md` diff
- Skim through sub-tasks in https://linear.app/osmosis/issue/DATA-174/[candidaterouteopt]-optimize-candidate-routes
- If unclear, rereview both doc files in full

PR #357, opened 2024-06-27
feat: Implement CandidateRouteSearchDataWorker

p0mvn:
Docs: https://github.com/osmosis-labs/sqs/pull/355

Task: https://linear.app/osmosis/issue/DATA-247/[candidaterouteopt]-implement-candidateroutesearchdataworker

- implements barebones candidate route search data worker
- wires it up in the worker/listener pipeline
- wires it up to healthcheck
- wires it up to router use case to persist the pre-computed data for routing

### DIFF
--- a/router/usecase/worker/candidate_route_search_data_worker.go
+++ b/router/usecase/worker/candidate_route_search_data_worker.go
@@ -1,0 +1,65 @@
+package worker
+
+import (
+	"context"
+
+	"github.com/osmosis-labs/sqs/domain"
+	"github.com/osmosis-labs/sqs/domain/mvc"
+	"github.com/osmosis-labs/sqs/log"
+	"github.com/osmosis-labs/sqs/sqsdomain"
+)
+
+type candidateRouteSearchDataWorker struct {
+	listeners                []domain.CandidateRouteSearchDataUpdateListener
+	poolsHandler             mvc.PoolHandler
+	candidateRouteDataHolder mvc.CandidateRouteSearchDataHolder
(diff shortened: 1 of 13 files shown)


router/usecase/worker/candidate_route_search_data_worker.go
```diff
@@ -0,0 +1,63 @@
...
+// OnPoolLiquidityCompute implements domain.PoolLiquidityComputeListener.
+func (c *candidateRouteSearchDataWorker) OnPoolLiquidityCompute(ctx context.Context, height uint64, blockPoolMetaData domain.BlockPoolMetadata) error {
+	// Compute search data and propagate error up the chain to fail the health check.
+	if err := c.ComputeSearchData(ctx, height, blockPoolMetaData); err != nil {
+		return err
+	}
+
+	// Notify listeners
+	for _, listener := range c.listeners {
+		_ = listener.OnSearchDataUpdate(ctx, height)
```
cryptomatictrader, line 42: Do we expect the listener will take long to finish? Might consider using separate thread.

cryptomatictrader (approved):
LGTM 👍

router/usecase/worker/candidate_route_search_data_worker.go
p0mvn: No, it is not expected. The only current use case is to have a health check run it.

I prefer introducing concurrency only when necessary for code simplicity, better debugging experience and bug resistance. 

This is already running in the background gorotutine so not blocking for the main flow.

PR #358, opened 2024-06-27
Fix: orderbook spotprice miscalculation

iboss-ptk:
- bump sqsdomain
- fix orderbook spot price miscalculation

### DIFF
--- a/router/usecase/pools/routable_cw_orderbook_pool.go
+++ b/router/usecase/pools/routable_cw_orderbook_pool.go
@@ -155,13 +155,15 @@ func (r *routableOrderbookPoolImpl) SetTokenOutDenom(tokenOutDenom string) {
 
 // CalcSpotPrice implements domain.RoutablePool.
 func (r *routableOrderbookPoolImpl) CalcSpotPrice(ctx context.Context, baseDenom string, quoteDenom string) (osmomath.BigDec, error) {
-	// Get the expected order direction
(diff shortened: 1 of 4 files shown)


router/usecase/pools/routable_cw_orderbook_pool.go
```diff
@@ -155,21 +155,28 @@ func (r *routableOrderbookPoolImpl) SetTokenOutDenom(tokenOutDenom string) {
...
 	tick := r.OrderbookData.Ticks[tickIdx]
 
 	// Calculate the price for the current tick
-	return clmath.TickToPrice(tick.TickId)
+	tickPrice, err := clmath.TickToPrice(tick.TickId)
+	if err != nil {
+		return osmomath.BigDec{}, err
+	}
+
+	return convertValue(osmomath.OneBigDec(), tickPrice, directionOut), nil
```
p0mvn: Can be precomputed since `converValue` does not use mutative math

p0mvn (approved):
Nice catch

router/usecase/pools/routable_cw_orderbook_pool.go
iboss-ptk: I'm not sure what you mean. Could you elaborate?

p0mvn: Define `var oneDec = osmomath.OneBigDec()` in global space and reference here

PR #390, opened 2024-07-15
Feat: swap Exact Amount Out, SQS Routing and API changes

deividaspetraitis:
WIP


router/delivery/http/router_handler.go
```diff
@@ -59,13 +63,98 @@ func NewRouterHandler(e *echo.Echo, us mvc.RouterUsecase, tu mvc.TokensUsecase,
 	e.GET(formatRouterResource("/state"), handler.GetRouterState)
 }
 
+// GetQuoteRequest represents swap quote request for the /router/quote endpoint.
+type GetQuoteRequest struct {
```
deividaspetraitis: Idea is to reduce HTTP handler logic to make it more readable  and cover HTTP layer with unit tests ( http respones, status codes ). Result is tha requests self contained requests that are capable to: 

- validate them selves
- unmarshal data from HTTP. 

Requests may live in separate API ( representation layer )  package.
- I was wondering whether such approach would be acceptable?

router/delivery/http/router_handler.go
```diff
@@ -59,13 +63,98 @@ func NewRouterHandler(e *echo.Echo, us mvc.RouterUsecase, tu mvc.TokensUsecase,
...
+func (r *GetQuoteRequest) IsSwapExactAmountIn() bool {
+	return r.TokenIn != nil && r.TokenOutDenom != ""
+}
+
+func (r *GetQuoteRequest) IsSwapExactAmountOut() bool {
+	return r.TokenOut != nil && r.TokenInDenom != ""
+}
+
+// TODO: it should NOT depend on echo context?
+func (r *GetQuoteRequest) Validate(c echo.Context) error {
```
deividaspetraitis: Should we make Requests routing library agnostic?

deividaspetraitis:
Hey @p0mvn, at this point this PR intruduces `tokenOut` and `tokenInDenom` API parameters and wires those to make calculation of exact amount out feature to work as expected, however I still need to adjust reponse according. I am also would like to propose  one idea how to decouple some logic from handlers in quest to make it more readable and cover with tests, can I ask for your input for that as well?

router/delivery/http/router_handler.go
```diff
@@ -85,19 +174,27 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
-		return c.JSON(domain.GetStatusCode(err), domain.ResponseError{Message: err.Error()})
+	// Validate the request
+	if err := req.Validate(c); err != nil {
+		return err
 	}
 
-	tokenOutDenom, tokenIn, err := getValidRoutingParameters(c)
-	if err != nil {
-		return err
+	var (
```
deividaspetraitis, line 94: I think it is more business logic than HTTP handlers. Should we pass `GetQuoteRequest` to `tokensusecase.GetOptimalQuote` and move this logic there?

p0mvn: Good point. 

I think we should keep `GetOptimalQuote` as is but rename it to `GetOptimalQuoteOutGivenIn`.

Then, add a new method`GetOptimalQuoteInGivenOut` to encapsulate the new business logic.

However, I think that the logic so far for deciding on the inputs makes sense to keep here. What I would add to `GetOptimalQuoteInGivenOut` is the logic for inverting the routes to display the final result.

We will likely also need a new implementation/handling logic for `PrepareResult`:
https://github.com/osmosis-labs/sqs/blob/16f75b936a18c3bf8ce2a1f85d3f542712c40b2c/router/usecase/quote.go#L39

router/delivery/http/router_handler.go
p0mvn: This is a good idea, and I'm happy to experiment. However, if we follow this pattern, I would love for us to define a plan to switch the remaining APIs onto the same as a follow-up and have tasks cut for that before merging 🙏 .

I think we would benefit from a general `types` package per module (i.e `router`, `pools`, `tokens`).  We could put the requests into such a package. Later on, we could also move on module-specific errors into the same location.

Such `types` approach would make us consistent with the chain logic where [every module](https://github.com/osmosis-labs/osmosis/tree/main/x/concentrated-liquidity) has a `types` package.

router/delivery/http/router_handler.go
p0mvn: Not sure what you mean by this question. Could you elaborate?

deividaspetraitis:
Hey @p0mvn, I have pushed PoC and as per my testing based on TDD spec  it outputs expected results. As we discussed earlier I have achieved this result with non-elagant approach and proposed more elagant one [here](https://github.com/osmosis-labs/sqs/pull/390/files#diff-43241936c83e05ea555d0b3116eb68c8df7d66d29e6d09bd4544cb67872ea406R226).

P.S. Sorry for failing pipeline, it's because I haven't implemented some methods for test mocks. Otherwise project builds and runs fine.

router/usecase/pools/routable_result_pool.go
```diff
@@ -51,6 +60,17 @@ func NewRoutableResultPool(ID uint64, poolType poolmanagertypes.PoolType, spread
 	}
 }
 
+func NewExactAmountOutRoutableResultPool(ID uint64, poolType poolmanagertypes.PoolType, spreadFactor osmomath.Dec, tokenInDenom string, takerFee osmomath.Dec, codeID uint64) domain.RoutablePool {
```
deividaspetraitis: TODO: remove

router/usecase/quote.go
```diff
@@ -96,6 +112,17 @@ func (q *quoteImpl) PrepareResult(ctx context.Context, scalingFactor osmomath.De
 	return q.Route, q.EffectiveFee, nil
 }
 
+func (q *QuoteExactAmountOut) PrepareResult(ctx context.Context, scalingFactor osmomath.Dec) ([]domain.SplitRoute, osmomath.Dec, error) {
+	for _, r := range q.Route {
+		for _, p := range r.GetPools() {
```
deividaspetraitis: Inverts pool input and output

router/delivery/http/router_handler.go
```diff
@@ -109,32 +218,64 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
+	// If a Quote is swap method specific, we can remove all below
+	if req.IsSwapExactAmountIn() {
+		return c.JSON(http.StatusOK, quote)
+	}
+
+	// For swap exact amount out, we reformat quote: invert amount in and amount out.
+	q := usecase.QuoteExactAmountOut{
+		AmountOut:               quote.GetAmountIn(),
+		AmountIn:                quote.GetAmountOut(),
+		Route:                   quote.GetRoute(),
```
p0mvn: Do we need to reformat the route as well?

router/usecase/pools/routable_balancer_pool.go
```diff
@@ -51,11 +67,23 @@ func (r *routableBalancerPoolImpl) ChargeTakerFeeExactIn(tokenIn sdk.Coin) (toke
 	return tokenInAfterTakerFee
 }
 
+// ChargeTakerFee implements domain.RoutablePool.
+// Charges the taker fee for the given token in and returns the token in after the fee has been charged.
+func (r *routableBalancerPoolImpl) ChargeTakerFeeExactOut(tokenOut sdk.Coin) (tokenOutAfterFee sdk.Coin) {
+	tokenOutAfterTakerFee, _ := poolmanager.CalcTakerFeeExactOut(tokenOut, r.TakerFee)
+	return tokenOutAfterTakerFee
+}
```
p0mvn: Why do we need this method?

The quote logic of "out given in" should handle the taker fee charge by charging it on token out

router/usecase/pools/routable_concentrated_pool.go
```diff
@@ -192,11 +193,136 @@ func (r *routableConcentratedPoolImpl) CalculateTokenOutByTokenIn(ctx context.Co
...
+// CalculateTokenInByTokenOut implements domain.RoutablePool.
+// It calculates the amount of token out given the amount of token in for a concentrated liquidity pool.
+// Fails if:
+// - the underlying chain pool set on the routable pool is not of concentrated type
+// - fails to retrieve the tick model for the pool
+// - the current tick is not within the specified current bucket range
+// - tick model has no liquidity flag set
+// - the current sqrt price is zero
+// - rans out of ticks during swap (token in is too high for liquidity in the pool)
+func (r *routableConcentratedPoolImpl) CalculateTokenInByTokenOut(ctx context.Context, tokenOut sdk.Coin) (sdk.Coin, error) {
```
p0mvn: Why do we need to implement this too? I thought we concluded that we can bypass implementing these

deividaspetraitis: This one was actually unused just to demonstrate how much extra code it would cost to implement rest of the relevant methods specific to the swap method. I have removed it.

router/usecase/pools/routable_balancer_pool.go
deividaspetraitis: Hmm, I must made a mistake when making a notes on my end, because I have noted that Taker fees needed to be calculated based on swap method. I have removed  `ChargeTakerFeeExactOut` methods from all pools.

router/delivery/http/router_handler.go
deividaspetraitis: Yes, we need to invert amounts, it's done [here](https://github.com/osmosis-labs/sqs/pull/390/files#diff-8e7615d9ce68d7727b44257f7c7c794396152b9f9d2afdd09505fbdb4880f18aR152)

router/usecase/quote.go
```diff
@@ -133,3 +156,63 @@ func (q *quoteImpl) String() string {
...
+
+// PrepareResult implements domain.Quote.
+// PrepareResult mutates the quote to prepare
+// it with the data formatted for output to the client.
+// Specifically:
+// It strips away unnecessary fields from each pool in the route.
+// Computes an effective spread factor from all routes.
+//
+// Returns the updated route and the effective spread factor.
+func (q *quoteExactAmountOut) PrepareResult(ctx context.Context, scalingFactor osmomath.Dec) ([]domain.SplitRoute, osmomath.Dec, error) {
```
deividaspetraitis, line 182: TODO: test

router/delivery/http/router_handler.go
deividaspetraitis: Thanks for sharing your ideas! I think it's a way to go then, I have moved GetQuoteRequest to `types` package and created [DATA-287](https://linear.app/osmosis/issue/DATA-287/encapsulate-requests) to address rest of the endpoints and requests.

deividaspetraitis:
Reopened in https://github.com/osmosis-labs/sqs/pull/397 to make it easier to follow.

PR #391, opened 2024-07-15
chore(ci): fix staging ci auto deploy pipeline

p0mvn:
Auto-deploy pipeline breaks staging deploys because it pushes a broken config that refers to node as localhost when it should be doing so by container name.

Fixed by adding a jq script to modify the breaking points.

Did not test this but 80% sure it will just work

## Summary by CodeRabbit

- **Chores**
	- Updated the Docker image deployment workflow for improved staging environment setup.

### DIFF
--- a/.github/workflows/push-dev-docker-images.yml
+++ b/.github/workflows/push-dev-docker-images.yml
@@ -69,8 +69,11 @@ jobs:
     - 
       if: ${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
       name: Send Webhook to deploy automatically to stage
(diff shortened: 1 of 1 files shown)


alessandrolomanto (approved):
This is theoretically not necessary because we make this substitution with ansible.
https://github.com/osmosis-labs/infrastructure/blob/main/environments/sqs-osmosis-zone/ansible/roles/prepare_local_folder/tasks/main.yaml#L32-L62
But I approve of it anyway because it's also good to have it here, it's better than ansible.

I tried the commands and they work fine, so looks good to me! Thank you!

PR #397, opened 2024-07-19
Feat: swap Exact Amount Out, SQS Routing and API changes

deividaspetraitis:
This pull request introduces getting a quote for token swap method Exact  Amount Out. Indroduced changes are based on existing Exact Amount In method in efforts to reuse available functionality and avoid duplicating efforts.

Additional improvements such as encapsulation of request for GetOptimalQuote is proposed.

## Summary by CodeRabbit

- **New Features**
	- Introduced new query parameters for token transactions, enhancing API flexibility.
	- Expanded API documentation to clarify input/output options for token transactions.

- **Bug Fixes**
	- Improved error handling related to token validation and swap methods.

- **Documentation**
	- Updated API specifications and documentation to reflect changes in query parameters and request structures.

- **Refactor**
	- Streamlined the request validation process and overall code structure for maintainability. 

- **Tests**
	- Introduced new test cases for validating quoting functionality based on exact amounts in and out.


router/usecase/quote.go
```diff
@@ -130,3 +153,63 @@ func (q *quoteImpl) String() string {
...
+
+// PrepareResult implements domain.Quote.
+// PrepareResult mutates the quote to prepare
+// it with the data formatted for output to the client.
+// Specifically:
+// It strips away unnecessary fields from each pool in the route.
+// Computes an effective spread factor from all routes.
+//
+// Returns the updated route and the effective spread factor.
+func (q *quoteExactAmountOut) PrepareResult(ctx context.Context, scalingFactor osmomath.Dec) ([]domain.SplitRoute, osmomath.Dec, error) {
```
deividaspetraitis, line 182: TODO: test

router/delivery/http/router_handler.go
```diff
@@ -109,23 +113,23 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
-
-	// Disable split routes if singleRoute is true
-	if isSingleRoute {
+	var routerOpts []domain.RouterOption
+	if req.SingleRoute {
 		routerOpts = append(routerOpts, domain.WithMaxSplitRoutes(domain.DisableSplitRoutes))
 	}
 
-	quote, err := a.RUsecase.GetOptimalQuote(ctx, tokenIn, tokenOutDenom, routerOpts...)
+	quote, err := a.RUsecase.GetOptimalQuote(ctx, *tokenIn, tokenOutDenom, req.SwapMethod(), routerOpts...)
```
p0mvn: Instead of passing in the swap method as a parameter, would it be possible to create a wrapper around `GetOptimalQuote` such as `GetOptimalQuoteInGivenOut` that would call `GetOptimalQuote` under the hood and reformat the result into the final type?

The reason I'm suggesting this is that passing in an argument to control the execution flow is a code smell that adds complexity.

By having a separate function, we would be able to functionally decouple the "in given out" implementation while also avoiding any changes to the core `GetSplitQuote`

p0mvn (commented):
Nice work! I have one primary recommendation/question

cryptomatictrader:
Nice work! LGTM 👍

router/delivery/http/router_handler.go
deividaspetraitis: Thanks for the feedback Roman, I have implemented [GetOptimalQuoteInGivenOut](https://github.com/osmosis-labs/sqs/pull/397/files#diff-9d1423fc5abbd62d4ff5f8151c5277b0dbc5af89a2e603f2b0c211835f690a11R216)!  :rocket:

domain/tokens.go
```diff
@@ -52,3 +52,18 @@ type TokenRegistryLoader interface {
 	// FetchAndUpdateTokens fetches tokens from the chain registry and updates the token registry.
 	FetchAndUpdateTokens() error
 }
+
+// SwapMethod is the type of token swap method.
+type TokenSwapMethod int
```
p0mvn, line 61: nit: could we make it a boolean if there are only 2 options?

router/delivery/http/router_handler.go
```diff
@@ -85,21 +81,21 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
-		return c.JSON(domain.GetStatusCode(err), domain.ResponseError{Message: err.Error()})
+	var req types.GetQuoteRequest
+	if err := UnmarshalRequest(c, &req); err != nil {
+		return err
 	}
 
-	tokenOutDenom, tokenIn, err := getValidRoutingParameters(c)
-	if err != nil {
+	// Validate the request
+	if err := req.Validate(); err != nil {
```
p0mvn, line 95: Nice! I like the clean-up

router/delivery/http/router_handler.go
```diff
@@ -109,23 +105,29 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
+		quote, err = a.RUsecase.GetOptimalQuoteInGivenOut(ctx, *req.TokenOut, req.TokenInDenom, routerOpts...)
+	}
+
 	if err != nil {
 		return err
 	}
 
 	scalingFactor := oneDec
-	if shouldApplyExponents {
+	if req.ApplyExponents {
```
p0mvn, line 136: This is a drive-by recommendation - we don't need this logic for applying exponents. Let's make a task to track its removal

router/delivery/http/router_handler.go
```diff
@@ -109,23 +105,29 @@ func (a *RouterHandler) GetOptimalQuote(c echo.Context) (err error) {
...
 		scalingFactor = a.getSpotPriceScalingFactor(tokenIn.Denom, tokenOutDenom)
 	}
 
+	//  PrepareResult requires swap method to be passed in, because based on that
+	// it needs to calculate takers fee for each pool in the route.
 	_, _, err = quote.PrepareResult(ctx, scalingFactor)
```
p0mvn: Not seeing the swap method passed - I'm guessing that the comment is now obsolete?

router/usecase/pools/routable_balancer_pool.go
```diff
@@ -20,6 +20,7 @@ var _ domain.RoutablePool = &routableBalancerPoolImpl{}
 
 type routableBalancerPoolImpl struct {
 	ChainPool     *balancer.Pool "json:\"pool\""
+	TokenInDenom  string         "json:\"token_in_denom\""
 	TokenOutDenom string         "json:\"token_out_denom\""
```
p0mvn: Let's add "omitempty" to both since we can have only one or the other for any quote

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -22,6 +22,7 @@ type routableAlloyTransmuterPoolImpl struct {
 	ChainPool           *cwpoolmodel.CosmWasmPool         "json:\"pool\""
 	AlloyTransmuterData *cosmwasmpool.AlloyTransmuterData "json:\"alloy_transmuter_data\""
 	Balances            sdk.Coins                         "json:\"balances\""
+	TokenInDenom        string                            "json:\"token_in_denom\""
 	TokenOutDenom       string                            "json:\"token_out_denom\""
```
p0mvn: ditto for all pools with omitempty

router/usecase/quote.go
```diff
@@ -13,6 +13,16 @@ import (
 	"github.com/osmosis-labs/osmosis/osmomath"
 )
 
+var (
+	one = osmomath.OneDec()
+)
+
+var (
+	_ domain.Quote = &quoteImpl{}
```
p0mvn: Let's rename this to `quoteExactAmountIn`

router/usecase/quote.go
```diff
@@ -130,3 +134,63 @@ func (q *quoteImpl) String() string {
...
+
+		// invert the in and out amounts for each pool
+		for _, p := range route.GetPools() {
+			p.SetTokenInDenom(p.GetTokenOutDenom())
+			p.SetTokenOutDenom("")
+		}
+	}
+
+	return q.Route, q.EffectiveFee, nil
+}
```
p0mvn: Consider moving this to a new file `quote_in_given_out.go` and also renaming `quote.go` to `quote_out_given_in.go`

router/usecase/quote.go
```diff
@@ -130,3 +134,63 @@ func (q *quoteImpl) String() string {
...
+	q.Route = q.quoteImpl.Route
+	q.EffectiveFee = q.quoteImpl.EffectiveFee
+	q.PriceImpact = q.quoteImpl.PriceImpact
+	q.InBaseOutQuoteSpotPrice = q.quoteImpl.InBaseOutQuoteSpotPrice
+
+	for i, route := range q.Route {
+		route, ok := route.(*RouteWithOutAmount)
+		if !ok {
+			panic("invalid route type") // should never happen
+		}
```
p0mvn: Let's bubble up the error rather than panicking so that a client gets a clearer message if this indeed occurs for one or the other reason

router/usecase/router_usecase.go
```diff
@@ -211,6 +211,24 @@ func (r *routerUseCaseImpl) GetOptimalQuote(ctx context.Context, tokenIn sdk.Coi
 	return finalQuote, nil
 }
 
+// GetOptimalQuoteInGivenOut returns an optimal quote through the pools for the exact amount out token swap method.
+// Underlying implementation is the same as GetOptimalQuote, but the returned quote is wrapped in a quoteExactAmountOut.
+func (r *routerUseCaseImpl) GetOptimalQuoteInGivenOut(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom string, opts ...domain.RouterOption) (domain.Quote, error) {
```
p0mvn, line 216: Let's add a "happy path" unit test for this.

router/usecase/quote_test.go
```diff
@@ -122,127 +123,334 @@ func (s *RouterTestSuite) TestPrepareResult() {
...
+					InAmount:  totalInAmount.QuoRaw(2),
+					OutAmount: totalOutAmount.QuoRaw(2),
 				},
 			},
-
-			InAmount:  totalInAmount.QuoRaw(2),
-			OutAmount: totalOutAmount.QuoRaw(2),
+			// (0.02 + (1 - 0.02) * 0.0004) * 0.5 + 0.003 * 0.5
+			expectedEffectiveFee: "0.011696000000000000",
+			expectedJSON: `{
```
p0mvn: Let's remove the `expectedJSON` because it is more difficult to maintain. Every time we make a change, we would have to modify these payloads which would be time-consuming.

Instead, we can assert similar guarantees more dynamically at the integration test level

p0mvn (commented):
Nice work!

Most of my comments are cosmetic.

I have the following questions:
- Custom direct quote implementation - what is our plan with that? Is there a task that also tracks its implementation for "in given out"
- Observability - do you think we need to add any new instrumentation, build dashboards or add alerts?
   * From my side, I think that we can split the `/router/quote` P95/99 dashboards to be by-type ("in given out" vs "out given in") 

I would like to also request to add integration tests before merge and confirm that they all pass. Something similar to [these](https://github.com/osmosis-labs/sqs/blob/bd8d5e2b8248bc7e9c6934390b42bb4bdbc0642b/tests/test_router_quote.py#L33-L117)

Prior to merge, could you please also add a load test for this endpoint ([ref](https://github.com/osmosis-labs/sqs/blob/v25.x/locust/locustfile.py))? And then sharing the results when all other endpoints are commented out/disabled in the load test

Please feel free to create integration/load tests in separate PRs made against the current branch. Once we have integration/load test results and all of the cosmetic comments are addressed, we should be good to merge!

router/delivery/http/router_handler.go
deividaspetraitis: Created a task [DATA-289](https://linear.app/osmosis/issue/DATA-289/sqs-remove-logic-for-applying-exponents) to address that.

router/delivery/http/router_handler.go
deividaspetraitis: Good catch! I have removed this comment.

router/usecase/pools/routable_balancer_pool.go
deividaspetraitis: Makes sense, good catch, I have updated every pool to reflect this suggestion, thanks!

router/usecase/quote.go
deividaspetraitis: To be honest initially I had the same idea, but decided not to do that to avoid making PR larger. It's updated now.

router/usecase/quote.go
deividaspetraitis: Splitted `quote.go` to ` quote_in_given_out.go` and `quote_out_given_in.go`

router/usecase/quote.go
deividaspetraitis: Adjusted tor return `types.ErrInvalidRouteType` in such scenario.

domain/tokens.go
deividaspetraitis: Technically there are 3 options, becaues one is used to indicate [not valid](https://github.com/osmosis-labs/sqs/pull/397/files#diff-3b0a84c19b354d3d3096aedf0c4f92bf939553fc83f59e20a229454bf149c4c4R68) swap method. Personaly I would keep as is, do you think it's worth giving a try simplifying to make 2 instead?

router/types/request.go
```diff
@@ -0,0 +1,101 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/labstack/echo/v4"
+)
+
+// GetQuoteRequest represents swap quote request for the /router/quote endpoint.
+type GetQuoteRequest struct {
+	TokenIn        *sdk.Coin
+	TokenOutDenom  string
+	TokenOut       *sdk.Coin
```
p0mvn, line 14: nit: we don't have to make coins be pointers. If `TokenIn`, for example, is uninitialized, we can check that with `TokenIn.IsNil()`

domain/tokens.go
p0mvn: Let's keep it 👍

pools/usecase/pools_usecase.go
```diff
@@ -312,7 +312,7 @@ func (p *poolsUseCase) StorePools(pools []sqsdomain.PoolI) error {
 		// If orderbook, update top liquidity pool for base and quote denom if it has higher liquidity capitalization.
 		sqsModel := pool.GetSQSPoolModel()
 		cosmWasmPoolModel := sqsModel.CosmWasmPoolModel
-		if cosmWasmPoolModel != nil && cosmWasmPoolModel.IsOrderbook() {
+		if cosmWasmPoolModel != nil && cosmWasmPoolModel.Data.Orderbook != nil && cosmWasmPoolModel.IsOrderbook() {
```
p0mvn: Oh wow - did this have a bug?

deividaspetraitis: Initially I thought it's because of outdated `osmosisd` local version, but I am getting same on `dev env` too. I probably would need to check this behaviour against `v25.x` branch, it's on my list.

p0mvn: I see - nice catch - let's separate it into its own PR for quick merge and deployment 🙏

router/usecase/quote_test.go
deividaspetraitis: For a reference:  This suggestion was accepted since [it is](https://github.com/osmosis-labs/sqs/pull/397/files#diff-997a6364dda024fd4d1a647f5de61deb04cd144f434ea04a0d6839dc68af9406R176) using same approach and test files as [handlers](https://github.com/osmosis-labs/sqs/pull/397/files#diff-847392b583119d093d0335703a7502a31c338dfac440da34d427cece264d0cf7R96).

pools/usecase/pools_usecase.go
deividaspetraitis: Created a separate [PR-411](https://github.com/osmosis-labs/sqs/pull/411) :rocket:

p0mvn (approved):
Great work!

Tested with swagger and left a UX request in DMs. 

Seeing that tests are progressing in separate PR, pre-approving this.

Let's get the swagger change in, resolved conflicts and merge this.

Additionally, I created [a task](https://linear.app/osmosis/issue/SQS-303/implement-in-given-out-for-custom-direct-quote) to similarly integrate `/custom-direct-quote` into "in given out". This is launch blocking.

deividaspetraitis:
> Great work!
> 
> Tested with swagger and left a UX request in DMs.
> 
> Seeing that tests are progressing in separate PR, pre-approving this.
> 
> Let's get the swagger change in, resolved conflicts and merge this.
> 
> Additionally, I created [a task](https://linear.app/osmosis/issue/SQS-303/implement-in-given-out-for-custom-direct-quote) to similarly integrate `/custom-direct-quote` into "in given out". This is launch blocking.

Conflicts were resolved, swagger updated, tests merged in re-run ( integration/unit/locust ) after conflicts resolution and it seems everything is working as expected.  For [SQS-303](https://linear.app/osmosis/issue/SQS-303/implement-in-given-out-for-custom-direct-quote)  separate PR will be opened. Merging in. :rocket: Thank you for your support! :pray:

PR #400, opened 2024-07-25
Feat: swap Exact Amount Out: Integration Tests

deividaspetraitis:
This Pull Request introduces integration tests for exact amount out quotes to make sure API functions as expected.

Tests for Exact Amount Out basically replicates ones of Exact Amount In and encapsulates those in its own test class. Initially goal was to came up with existing tests reuse strategy, but I was not able to find elegant approach at this point.

For clarity old test class was renamed to reflect it's scope.

---

Error in the tests are not related. Should we create a task to fix and track them?

```
============================================= short test summary info =============================================
FAILED tests/test_candidate_routes.py::TestCandidateRoutes::test_all_valid_tokens[http://localhost:9092-ibc/0F9E927
7B61A78CB31014D541ACA5BF6AB06DFC4524C4C836490B131DAAECD78] - AssertionError: Error: no routes found for token in ib
c/0F9E9277B61A78CB31014D541ACA5BF6AB06DFC4524C4C836490B1...
FAILED tests/test_pools.py::TestPools::test_canonical_orderbook[http://localhost:9092] - AssertionError: Canonical
orderbooks are None
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[http://localhost:9092-ibc/56C276FC1
36E239449DCE664292DBEEF5795C4EF4B5B35DB98BD1C0948274867] - AssertionError: ibc/56C276FC136E239449DCE664292DBEEF5795
C4EF4B5B35DB98BD1C0948274867, SQS price is zero
=================================== 3 failed, 1964 passed, 8 skipped in 17.92s ====================================

```

### DIFF
--- a/tests/quote.py
+++ b/tests/quote.py
@@ -2,7 +2,7 @@ import conftest
 
 class Quote:
     @staticmethod
-    def choose_error_tolerance(amount_in: int):
+    def choose_error_tolerance(amount: int):
          # This is the max error tolerance of 7% that we allow.
         # Arbitrarily hand-picked to avoid flakiness.
         error_tolerance = 0.07
@@ -10,11 +10,11 @@ class Quote:
         # flakiness. Therefore, we increase the error tolerance based on the amount in swapped.
         # The values are arbitrarily hand-picked and can be adjusted if necessary.
         # This seems to be especially relevant for the Astroport PCL pools.
-        if amount_in > 60_000_000_000:
+        if amount > 60_000_000_000:
             error_tolerance = 0.16
-        elif amount_in > 30_000_000_000:
+        elif amount > 30_000_000_000:
             error_tolerance = 0.13
-        elif amount_in > 10_000_000_000:
+        elif amount > 10_000_000_000:
             error_tolerance = 0.10
 
         return error_tolerance
--- a/tests/quote.py
+++ b/tests/quote.py
@@ -1,0 +1,57 @@
+import conftest
+
+class Quote:
+    @staticmethod
+    def choose_error_tolerance(amount_in: int):
+         # This is the max error tolerance of 7% that we allow.
+        # Arbitrarily hand-picked to avoid flakiness.
+        error_tolerance = 0.07
+        # At a higher amount in, the volatility is much higher, leading to
+        # flakiness. Therefore, we increase the error tolerance based on the amount in swapped.
+        # The values are arbitrarily hand-picked and can be adjusted if necessary.
+        # This seems to be especially relevant for the Astroport PCL pools.
+        if amount_in > 60_000_000_000:
+            error_tolerance = 0.16
+        elif amount_in > 30_000_000_000:
+            error_tolerance = 0.13
+        elif amount_in > 10_000_000_000:
+            error_tolerance = 0.10
+
+        return error_tolerance
+
+    @staticmethod
+    def is_transmuter_in_single_route(routes):
+        """
+        Returns true if there is a single route with
+        one transmuter pool in it.
+        """
+        if len(routes) == 1 and len(routes[0].pools) == 1:
+            pool_in_route = routes[0].pools[0]
+            pool = conftest.shared_test_state.pool_by_id_map.get(str(pool_in_route.id))
+            e2e_pool_type = conftest.get_e2e_pool_type_from_numia_pool(pool)
+
+            return  e2e_pool_type == conftest.E2EPoolType.COSMWASM_TRANSMUTER_V1
+        
+        return False
+
+    @staticmethod
+    def validate_fee(quote):
+        """
+        Validates fee returned in the quote response.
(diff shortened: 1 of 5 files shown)


tests/test_router_quote_out_given_in.py
```diff
@@ -0,0 +1,243 @@
...
+# The choice is arbitrary and was made based on testing at the time of creation.
+# In the future, we might lower or increase this value based on the performance of the system.
+HIGH_LIQ_PRICE_IMPACT_CHECK_USD_AMOUNT_IN_THRESHOLD = 5000
+
+# The max price impact threshold for the high liquidity check
+HIGH_LIQ_MAX_PRICE_IMPACT_THRESHOLD = 0.5
+
+# Test suite for the /router/quote endpoint
+# Test runs tests for exact amount out quotes.
+class TestExactAmountOutQuote:
```
p0mvn: nit:I think the file name should be invested "out_given_in" to "in_given_out"

tests/test_router_quote_out_given_in.py
```diff
@@ -0,0 +1,243 @@
...
+        out_base_usd_quote_price = conftest.get_usd_price_scaled(token_out_denom)
+        expected_out_base_in_quote_price = out_base_usd_quote_price / in_base_usd_quote_price 
+
+        # Compute expected token out
+        expected_token_in = int(amount_str) * expected_out_base_in_quote_price
+
+        token_out_amount_usdc_value = in_base_usd_quote_price * amount_out
+
+        # Chose the error tolerance based on amount in swapped.
+        error_tolerance = Quote.choose_error_tolerance(token_out_amount_usdc_value)
```
p0mvn: Nice - clean approach

p0mvn (commented):
Nice work. This looks good to me on the high-level.

I will spend some more time running tests tomorrow before approving

tests/test_router_quote_out_given_in.py
deividaspetraitis: Good catch! It's updated now.  :+1:

tests/quote.py
```diff
@@ -0,0 +1,57 @@
+import conftest
+
+class Quote:
+    @staticmethod
+    def choose_error_tolerance(amount_in: int):
```
p0mvn: nit: we call it on amount_out as well. Let's rename to a more general `amount`?

tests/test_router_quote_in_given_out.py
```diff
@@ -0,0 +1,243 @@
...
+
+        denom_in_data = conftest.shared_test_state.chain_denom_to_data_map.get(denom_in)
+        denom_in_precision = denom_in_data.get("exponent")
+        
+        # Compute spot price scaling factor.
+        spot_price_scaling_factor = Decimal(10)**6 / Decimal(10)**denom_in_precision
+
+        # Compute expected spot prices
+        out_base_in_quote_price = Decimal(denom_in_data.get("price"))
+        expected_out_base_in_quote_price = 1 / out_base_in_quote_price
```
p0mvn: Should `expected_out_base_in_quote_price` be `expected_in_base_out_quote_price` because we inverted the original `out_base_in_quote_price`?

tests/test_router_quote_in_given_out.py
```diff
@@ -0,0 +1,243 @@
...
+        assert response.status_code == expected_status_code, f"Error: {response.text}"
+        assert expected_latency_upper_bound_ms > elapsed_time_ms, f"Error: latency {elapsed_time_ms} exceeded {expected_latency_upper_bound_ms} ms, token in {token_in} and token out {token_out}" 
+
+        response_json = response.json()
+
+        # Return route for more detailed validation
+        return QuoteExactAmountOutResponse(**response_json)
+
+    def validate_quote_test(self, quote, expected_amount_out_str, expected_denom_out, spot_price_scaling_factor, expected_out_base_in_quote_price, expected_token_in, error_tolerance):
+        """
```
p0mvn: I know that this did not exist for "out given in" but, given new functionality, I think it would be helpful to validate the route format.

That is, that by following the returned route from the API, we go through the pools that indeed contain the specified tokens and end up correctly in the token in.

Additionally, since route pool may now contain either "token_in_denom" or "token_out_denom", this validation would be helpful.

I'm happy if you would like to add it in a separate PR.

p0mvn (approved):
Great work!

Pre-approving pending comments being addressed.

Additionally, would like to recommend expanding the route to validation (but can be done in a follow-up PR) if that's easier.

tests/quote.py
deividaspetraitis: Indeed, good catch!

tests/test_router_quote_in_given_out.py
deividaspetraitis: Good catch, updated! :+1:

deividaspetraitis:
Merging. Expanding the route vaildation will be done on separate PR.

PR #399, opened 2024-07-24
TestGetOptimalQuoteInGivenOut: Tests

deividaspetraitis:
This PR covers `TestGetOptimalQuoteInGivenOut` with tests.

As a side effect `GetOptimalQuote` method is broken is smaller parts to in order to be able to mock it. Such approach allows to tests smaller units indepently as well.

I have tested indroduced changes with Unit tests and integration tests:

```
============================================= short test summary info ==============================================
FAILED tests/test_candidate_routes.py::TestCandidateRoutes::test_all_valid_tokens[http://localhost:9092-ibc/0F9E9277
B61A78CB31014D541ACA5BF6AB06DFC4524C4C836490B131DAAECD78] - AssertionError: Error: no routes found for token in ibc/
0F9E9277B61A78CB31014D541ACA5BF6AB06DFC4524C4C836490B13...
FAILED tests/test_pools.py::TestPools::test_canonical_orderbook[http://localhost:9092] - AssertionError: Canonical o
rderbooks are None
FAILED tests/test_tokens_prices.py::TestTokensPrices::test_low_liq_token_prices[http://localhost:9092-ibc/56C276FC13
6E239449DCE664292DBEEF5795C4EF4B5B35DB98BD1C0948274867] - AssertionError: ibc/56C276FC136E239449DCE664292DBEEF5795C4
EF4B5B35DB98BD1C0948274867, SQS price is zero
==================================== 3 failed, 1490 passed, 8 skipped in 31.31s ====================================

```

This PR will be merged is `DATA-276` branch.

## Summary by CodeRabbit

- **New Features**
	- Introduced new parameters for token transactions, enhancing API flexibility.
	- Added methods for managing token denominations in various pools.
	- Implemented new quote retrieval methods for improved routing functionality.

- **Bug Fixes**
	- Refined error handling across the router, improving clarity and maintainability.

- **Documentation**
	- Updated API documentation to reflect changes in query parameters and error handling.

- **Tests**
	- Expanded test suite for the Router functionality, including new test cases for optimal quote retrieval.

- **Chores**
	- Enhanced code readability and organization across multiple files.

PR #403, opened 2024-07-26
Test get optimal quote in given out mainnet

deividaspetraitis:
This pull request adds `TestGetOptimalQuoteInGivenOut_Mainnet` which is basically a copy of [TestGetOptimalQuote_Mainnet](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/optimized_routes_test.go#L555) but makes calls to `GetOptimalQuoteInGivenOut`. At this point there is no simple way to construct `usecase.QuoteExactAmountOut`, this issue arrises in [TestGetOptimalQuote_Mainnet](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/optimized_routes_test.go#L555) too.


deividaspetraitis:
@p0mvn, I am not sure whether this PR should introduce capability to construct quotes for mainnet testing?  That seems quite demanding task. By constructing quotes I refer to capability to construct "expected" quote data structure to be compared with returned, same as [here](https://github.com/osmosis-labs/sqs/pull/399/files#diff-bd9a0212db199df7805c9f59694adca3801a281ca208fcee67afa5e0d1a9e4f9R713).  Without such capability current test cases has very small coverage.

Or should we leave as it is and just create a tracking task?

router/usecase/router_usecase_test.go
```diff
@@ -589,6 +589,123 @@ func (s *RouterTestSuite) TestConvertRankedToCandidateRoutes() {
...
+
+			// We only validate that error does not occur without actually validating the quote.
+			s.Require().NoError(err)
+
+			// TODO: update mainnet state and validate the quote for each test stricter.
+			quoteRoutes := quote.GetRoute()
+			s.Require().Len(quoteRoutes, tc.expectedRoutesCount)
+
+			// Validate that the quote is not nil
+			s.Require().NotNil(quote.GetAmountOut())
```
p0mvn: I know this was copied form "out given in" but one piece of validation that we could add is that the route is valid.


That is, by following the pools in the returned route, we can assert that the sequence is appropriate and tokens in and out exist in the relevant pools lover the route.

p0mvn:
> @p0mvn, I am not sure whether this PR should introduce capability to construct quotes for mainnet testing? That seems quite demanding task. By constructing quotes I refer to capability to construct "expected" quote data structure to be compared with returned, same as [here](https://github.com/osmosis-labs/sqs/pull/399/files#diff-bd9a0212db199df7805c9f59694adca3801a281ca208fcee67afa5e0d1a9e4f9R713). Without such capability current test cases has very small coverage.
> 
> Or should we leave as it is and just create a tracking task?

I don't think asserting on specific quote values would offer good enough dev UX due to liquidity values changing inter-block.

The way this test proves its usefulness is that for all specified token pairs, there are no errors, and at least one route returned with a non-zero amount.

Having such a high-level heuristic will already help us to have more confidence in the system while not having to be bottlenecked on updating the exact expected quote amounts each time we do a mainnet state update.

Math correctness should be ensured at a lower level of abstraction, and we can treat this particular test as more functional.

So, in terms of the next steps, I think it would be great if you could add the route validation I suggested in another comment, and then we can merge it 👍

router/usecase/router_usecase_test.go
deividaspetraitis: In terms of testing route/pools, let's consider that we are extending [TestGetOptimalQuote_Mainnet](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/optimized_routes_test.go#L555) instead for now. I may be missing something important here, but it seems that discussed validation will be not possible to implement since GetTokenInDenom is not set for the pools, - it is always empty quotes. That is visible in the [response](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/routertesting/parsing/quote_amount_in_response.json#L1) for the clients as well.
Now,  back to the [TestGetOptimalQuoteInGivenOut_Mainnet](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/router_usecase_test.go#L594), since GetOptimalQuoteInGivenOut is just a wrapper over quoteExactAmountIn call to GetTokenInDenom still returns empty string. Even if [response](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/routertesting/parsing/quote_amount_out_response.json#L1)  contains token_in_denom value we don't have it in the underlying data structures, because it's actually  token_out_denom , - result of inversion we apply during [PrepareResult](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/quote_out_given_in.go#L65) call. Note that this method is already [covered](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/quote_test.go#L40) with tests.
I suspect this the reason why other tests such as [ValidateRouterPools](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/routertesting/suite.go#L308) does not implement such test coverage either.  Please let me know whether I am missing something here or it's actually not feasible at this point?

deividaspetraitis:
Instead of duplicating code and efforts I have added route test and extended ` TestGetOptimalQuote_Mainnet` to support both quote types. I decided to keep current test name as is, IMO it still reflects test case intent quite well.

router/usecase/optimized_routes_test.go
```diff
@@ -652,18 +652,53 @@ func (s *RouterTestSuite) TestGetOptimalQuote_Mainnet() {
...
-			quote, err := mainnetUseCase.Router.GetOptimalQuote(context.Background(), sdk.NewCoin(tc.tokenInDenom, tc.amountIn), tc.tokenOutDenom)
+			// Setup the quotes to be tested
+			quotes := []func() (domain.Quote, error){
+				func() (domain.Quote, error) {
+					return mainnetUseCase.Router.GetOptimalQuote(context.Background(), sdk.NewCoin(tc.tokenInDenom, tc.amountIn), tc.tokenOutDenom)
+				},
+				func() (domain.Quote, error) {
+					return mainnetUseCase.Router.GetOptimalQuoteInGivenOut(context.Background(), sdk.NewCoin(tc.tokenInDenom, tc.amountIn), tc.tokenOutDenom)
+				},
+			}
```
p0mvn: I propose sharing the `tests` slice into the global space but still having separate test fixtures for "in given out" and "out given in" for readability.

Additionally, I don't think the current test for "in given out" is very useful because we essentially provide the same "in given out" quote under the hood.

As a result, for "in given out", I propose also calling:
```
_, _, err = quote.PrepareResult(context.Background(), osmomath.OneDec())
s.Require().NoError(err)
```

So that we can validate the reformatted route

deividaspetraitis: Do you think moving tests slice into the global space would be sufficient having in mind that `PrepareResult` is already covered with tests [here](https://github.com/osmosis-labs/sqs/blob/e1dcc8c6dc487293c8787e1ba50e4c72b7e48d60/router/usecase/quote_test.go#L40) with tests?

p0mvn: I think the goal of `TestPrepareResult` is different.

Here, we are trying to test that "in given out" quote functions off of mainnet state, acting more of a functional test.

To validate that properly, we should call `PrepareResult` because that's what converts "out given in" quote to "in given out"

p0mvn (approved):
GG 🚀

p0mvn:
Let's lint and merge this

deividaspetraitis:
Lints addressed with [b9bfe31](https://github.com/osmosis-labs/sqs/pull/403/commits/b9bfe312a79c654dd54544258389d25548018847), all checks have passed. Merging. :rocket:

PR #406, opened 2024-07-28
Mitigate the alloy rebalancing router breakage by invalidating route cache if all routes fail and ordering denoms by balances

p0mvn:
We encountered a production failure where an allBTC pool became imbalanced, causing disruptions in some allBTC routes.

There is a mechanism that silently ignores failing routes, which led us to conclude that all routes had failed.

The issue arose because all ranked and candidate routes were computed and cached to go through WBTC, which lost its liquidity.

Due to the limitation of having only 20 candidate routes, no other allBTC token was considered.

To address this, we now rank pool denominations by their balances. This means that if an alloyed pool goes out of balance, the recomputed candidate routes will first consider the tokens with the most liquidity. This change does not account for token precision but should still effectively handle the identified problem.

Additionally, we now invalidate the route caches for the token pair if all cached ranked routes fail to be computed

The change is unit and integration tested

## Summary by CodeRabbit

- **New Features**
	- Enhanced routing logic for better cache management and error handling.
	- Introduced flexibility in mock implementations for testing routing behaviors.
	- Added support for a new token identifier, expanding asset capabilities.
	- Broadened test coverage with new conversion scenarios for ALLBTC to TIA.

- **Bug Fixes**
	- Improved handling of balances for better liquidity management in routing.

- **Tests**
	- Updated tests to reflect changes in routing logic, expected outcomes, and mock behaviors, ensuring accurate validation of system functionality.


router/usecase/candidate_routes_test.go
```diff
@@ -126,20 +126,25 @@ func (s *RouterTestSuite) TestCandidateRouteSearcher_HappyPath() {
...
+		// {
+		// 	name:          "UMEE -> AKT",
+		// 	tokenIn:       sdk.NewCoin(UMEE, one),
+		// 	tokenOutDenom: AKT,
+		// },
+		// {
+		// 	name:          "ALLBTC -> USDC",
+		// 	tokenIn:       sdk.NewCoin(ALLBTC, one),
+		// 	tokenOutDenom: USDC,
+		// },
```
czarcas7ic: Why are these test cases no longer valid?

p0mvn: Thanks for catching it - was an accident. Restored now

domain/mocks/route_mock.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type RouteMock struct {
+	CalculateTokenOutByTokenInFunc      func(ctx context.Context, tokenIn types.Coin) (types.Coin, error)
+	ContainsGeneralizedCosmWasmPoolFunc func() bool
+	GetPoolsFunc                        func() []domain.RoutablePool
+	GetTokenOutDenomFunc                func() string
+	PrepareResultPoolsFunc              func(ctx context.Context, tokenIn types.Coin) ([]domain.RoutablePool, math.LegacyDec, math.LegacyDec, error)
+	StringFunc                          func() string
+}
```
p0mvn, line 18: Note: I ended up not using the mock but I think it is still useful for the future

ingest/usecase/ingest_usecase.go
```diff
@@ -473,6 +478,22 @@ func processSQSModelMut(sqsModel *sqsdomain.SQSPool) error {
 		newPoolDenoms = append(newPoolDenoms, denom)
 	}
 
+	// Sort the pool denoms by balance amount
+	// This is useful for edge case handling for certain pools such as alloyed
+	// where the token amounts might get imbalanced, making the liqudiity of one denom completely zero.
```
czarcas7ic:
```suggestion
	// where the token amounts might get imbalanced, making the liquidity of one denom completely zero.
```

ingest/usecase/ingest_usecase.go
```diff
@@ -473,6 +478,22 @@ func processSQSModelMut(sqsModel *sqsdomain.SQSPool) error {
...
+	// In that case, we would like to deprioritize the out-of-liquidity denoms.
+	sort.Slice(newPoolDenoms, func(i, j int) bool {
```
czarcas7ic, line 485: I think I am missing some ctx on how ordering the denoms by balance helps solve the de-prioritization

router/usecase/optimized_routes.go
```diff
@@ -61,6 +50,19 @@ func estimateAndRankSingleRouteQuote(ctx context.Context, routes []route.RouteIm
...
+		// This might be helpful in alloyed cases where the pool gets imbalanced and runs out of liquidity.
+		// If the original routes were computed only through the zero liquidity token, they will be recomputed
+		// through another token due to changed order.
+
+		// Note: the zero length check occurred at the start of function.
+		tokenOutDenom := routes[0].GetTokenOutDenom()
+
+		r.candidateRouteCache.Delete(formatCandidateRouteCacheKey(tokenIn.Denom, tokenOutDenom))
+		tokenInOrderOfMagnitude := GetPrecomputeOrderOfMagnitude(tokenIn.Amount)
+		r.rankedRouteCache.Delete(formatRankedRouteCacheKey(tokenIn.Denom, tokenOutDenom, tokenInOrderOfMagnitude))
```
czarcas7ic, line 64: I assume there is some logic somewhere that repopulates if the cache is empty

czarcas7ic (approved):

ingest/usecase/ingest_usecase.go
p0mvn: Our candidate route algorithm does BFS by goring through the pool denoms in order.

There is a limit of 20 candidate routes.

The bug was that all 20 candidate routes only container WBTC that had no liquidity.

With this idea, if liquidity changes, the denom order will change. Since cache also gets invalidated, we would recompute the routes where a higher liquidity token is first, as a result, reconstructing new 20 valid routes

router/usecase/optimized_routes.go
p0mvn: Yes, since cache is empty, it would try to recompute the routes and fill in the cache with the new routes

ingest/usecase/ingest_usecase.go
czarcas7ic: Ah, I understand now, thanks

PR #409, opened 2024-07-29
progress for quotes

p0mvn:


ingest/usecase/plugins/orderbookfiller/ordebook_filler_ingest_plugin.go
```diff
@@ -38,11 +43,86 @@ func New(poolsUseCase mvc.PoolsUsecase, routerUseCase mvc.RouterUsecase, tokensU
...
+		return err
+	}
+	baseOptimalQuote := quoteInOptimalQuote.GetAmountOut()
+	quoteOptimalQuote := quoteInOptimalQuote.GetAmountIn()
+
+	// Get execution price
+	optimalQuoteExecutionPrice := baseOptimalQuote.ToLegacyDec().Quo(quoteOptimalQuote.Amount.ToLegacyDec())
+
+	errTolerance := osmomath.ErrTolerance{
+		MultiplicativeTolerance: sdk.NewDecWithPrec(5, 2),
```
AlpinYukseloglu, line 124: Is this enough to account for error in `optimalQuoteExecutionPrice` due to e.g. not factoring in gas fee estimates?

p0mvn: Will try increasing but Adam's also added logic to account for gas fee

PR #416, opened 2024-07-30
Add prefetcher

ValarDragon:


sqsutil/prefetcher/prefetch.go
```diff
@@ -0,0 +1,94 @@
...
+
+	for range p.timer.C {
+		p.prefetch()
+	}
+}
+
+func (p *Prefetcher[T]) prefetch() {
+	newValue := p.updateFn()
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
```
p0mvn: For data that has many key/value pairs and frequent access to various keys, would we have to maintain a separate copy under `sync.Map` somewhere so that we don't cause latency spikes when acquiring a mutex here?

Or maybe should we have a separate prefetcher implementation as a `sync.Map` for this access pattern?

p0mvn (approved):

sqsutil/prefetcher/prefetch.go
ValarDragon: this should always be super quick, right?

This latency is the time it takes to either:
- copy a struct
- Swap a pointer (e.g. Slice or Map case)

So if you were querying a map, the overhead would be very small

p0mvn: Yes, but situation changes under load. I've had production use cases where with moderate write load and heavy ready load this combination of map and RW lock would cause meaningful RPS degradation and tail latency spikes. `sync.Map` would perform much better in those cases.

I am, however, in agreement to not over-focus on this at this stage and observe the patterns once we start load testing and profiling

ValarDragon: If this is true, we just switch to a CAS instruction

p0mvn: SGTM. I'm merging this

PR #421, opened 2024-07-31
fix bug applying invalid pool liquidity filter in the liquidity pricing pipeline.

p0mvn:
With the recent liquidity filter change, we accidentally changed the semantics of what it means to request an empty pool ID slice.

The original invariant meant empty pool IDs = no pools / no-op/

The new filter => empty pool IDs => all pools.

This broke the liquidity pricing pipeline. When no pools were modified in a block, we would provide an empty pool IDs slice to `GetPools`. This would return all pools with tokens that did not have prices in that block.

Therefore, we would mistakenly override pool liquidity capitalization with error that would completely nuke the router filtering.

## Testing

This was not caught by the integration test suite ran pre-merge likely because there was a series of blocks with modified pools.

However, overnight in staging a number of liquidity overrides happened for pools, exposing the bug. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with invalid pool liquidity filters affecting pricing calculations, improving the reliability of the liquidity management system.

- **New Features**
  - Introduced a `HadEmptyFilter` option to optimize pool fetching by avoiding unnecessary data retrieval when no specific pool IDs are provided.

- **Tests**
  - Added a test case to validate the behavior of pool filtering when an empty pool ID filter is applied, enhancing the robustness of the functionality.

### DIFF
--- a/pools/usecase/pools_usecase.go
+++ b/pools/usecase/pools_usecase.go
@@ -296,10 +296,15 @@ func (p *poolsUseCase) GetPools(opts ...domain.PoolsOption) ([]sqsdomain.PoolI, 
 	options := domain.PoolsOptions{
 		MinPoolLiquidityCap: 0,
 		PoolIDFilter:        []uint64{},
+		HadEmptyFilter:      false,
 	}
 
 	for _, opt := range opts {
 		opt(&options)
+	}
+
+	if options.HadEmptyFilter {
+		return nil, nil
 	}
 
 	var (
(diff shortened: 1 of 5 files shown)


pools/usecase/pools_usecase.go
```diff
@@ -316,7 +321,7 @@ func (p *poolsUseCase) GetPools(opts ...domain.PoolsOption) ([]sqsdomain.PoolI,
 			}
 
 			// Check filter is non-zero to avoid more expensive get liquidity cap check.
-			if pool.GetLiquidityCap().Uint64() >= options.MinPoolLiquidityCap {
+			if options.MinPoolLiquidityCap == 0 || pool.GetLiquidityCap().Uint64() >= options.MinPoolLiquidityCap {
```
p0mvn, line 324: Seems that Paddy's bug stemmed from this while testing v26.x

I could not repro the panic on v25.x but I added the safeguard just in case.

PR #438, opened 2024-08-09
SQS-306 | Rewrite CalcExitCFMM pool in SQS

deividaspetraitis:
Initial commit.

This PR attemps to rewrite `CalcExitCFMM` in order to get it optimised for SQS usecase by converting `cosmos-sdk` math data types to floats.

Locust load during profiling:

image

- run 2 represents before optimisation ( branch: `v25.x` )
- run 3 represents after optimisation ( branch `SQS-306` )

All graphs were generated with the same profiling command:

```bash
go tool pprof -http=:8080 http://localhost:9092/debug/pprof/profile?seconds=30
```

Flamegraph before:

image

Flamegraph After:

image

Result: load time is cut 30% with the current optimisation strategy.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new function for calculating exit pool amounts, enhancing exit pool management.
	- Updated existing method for improved clarity and performance in exit calculations.
	- Added comprehensive tests for exit calculations in different liquidity pool scenarios.

- **Bug Fixes**
	- Improved error handling in exit calculations to prevent over-exit scenarios.

- **Tests**
	- Added a new test case for validating exit amounts for various liquidity pools.

- **Chores**
	- Reorganized import statements for better readability.


pools/usecase/pools_usecase_test.go
```diff
@@ -566,6 +569,135 @@ func (s *PoolsUsecaseTestSuite) TestCalcExitCFMMPool_HappyPath() {
 	s.Require().False(actualCoins.Empty())
 }
 
+func (s *PoolsUsecaseTestSuite) TestCalcExitPool() {
```
deividaspetraitis, line 576: Ported from [lp_test.go](https://github.com/osmosis-labs/osmosis/blob/fde1776476d9c2f849dcbfb30ca3ec64d0e12863/x/gamm/pool-models/internal/cfmm_common/lp_test.go)

pools/usecase/pools_usecase_test.go
```diff
@@ -566,6 +569,135 @@ func (s *PoolsUsecaseTestSuite) TestCalcExitCFMMPool_HappyPath() {
...
+			exitingShares: twoAssetPool.GetTotalShares().AddRaw(1),
+			expError:      true,
+		},
+		{
+			name:          "three-asset pool, exiting shares grater than total shares",
+			pool:          &threeAssetPool,
+			exitingShares: threeAssetPool.GetTotalShares().AddRaw(1),
+			expError:      true,
+		},
+		// {
```
deividaspetraitis: Those fails and needs to be fixed

p0mvn:
Consider putting a side-by-side comparison of CPU flame graphs between before and after when profiling the same endpoint for the same amount of time. Seeing the proportion of CPU time spent in the relevant method would help to confirm the success of the reimplementation.

deividaspetraitis:
Good point @p0mvn! I have updated RP description to include points you have mentioned. Does it look right?

From the graphs it seems there a several places for improvement:

- math.LegacyDec
- int.ToLegacyDec
- GetTotalPoolLiquidity

I am not sure how we can optimise the last one.

deividaspetraitis:
With most recent commit optimised further 

- math.LegacyDec for `totalShares`
- int.ToLegacyDec for `existingShares`

PR description was updated to include the latest profiling graphs

pools/usecase/pools_usecase.go
```diff
@@ -540,7 +546,78 @@ func (p *poolsUseCase) CalcExitCFMMPool(poolID uint64, exitingShares osmomath.In
...
+	return calcExitPool(sdk.Context{}, pool, exitingSharesIn, exitFee)
+}
+
+// errMsgFormatSharesLargerThanMax is the error message format for when the exiting shares are larger than the max allowed.
+const errMsgFormatSharesLargerThanMax = "cannot exit all shares in a pool. Attempted to exit %f shares, max allowed is %f"
+
+// calcExitPool is a helper function to calculate the exit pool.
+// It is a direct port of the CalcExitPool function from the Gamm module with some modifications
+// to optimize the calculation for performance.
+func calcExitPool(ctx sdk.Context, pool types.CFMMPoolI, exitingSharesIn osmomath.Int, exitFee osmomath.Dec) (sdk.Coins, error) {
```
p0mvn: nit: let's link the gamm function here

pools/usecase/pools_usecase.go
```diff
@@ -540,7 +546,78 @@ func (p *poolsUseCase) CalcExitCFMMPool(poolID uint64, exitingShares osmomath.In
...
+	totalShares, err := strconv.ParseFloat(pool.GetTotalShares().String(), 64)
+	if err != nil {
+		return sdk.Coins{}, err
+	}
+
+	exitingShares, err := strconv.ParseFloat(exitingSharesIn.String(), 64)
+	if err != nil {
+		return sdk.Coins{}, err
+	}
```
p0mvn, line 569: I initially thought that we could convert to int64 and then to float for perf reasons but realizing that we would easily overflow the int

pools/usecase/pools_usecase.go
```diff
@@ -540,7 +546,78 @@ func (p *poolsUseCase) CalcExitCFMMPool(poolID uint64, exitingShares osmomath.In
...
+
+	shareOutRatio := refundedShares / totalShares
+
+	// exitedCoins = shareOutRatio * pool liquidity
+	poolLiquidity := pool.GetTotalPoolLiquidity(ctx)
+	exitedCoins := make(sdk.Coins, 0, len(poolLiquidity))
+
+	for _, asset := range poolLiquidity {
+		// round down here, due to not wanting to over-exit
+		amount := float64(asset.Amount.BigInt().Int64())
```
p0mvn: I think this might overflow and panic - we should consider using string conversation like you've done in another example

pools/usecase/pools_usecase.go
```diff
@@ -540,7 +546,78 @@ func (p *poolsUseCase) CalcExitCFMMPool(poolID uint64, exitingShares osmomath.In
...
+		}
+
+		if exitAmt >= amount {
+			return sdk.Coins{}, errors.New("too many shares out")
+		}
+
+		// If the exit amount is within 1e-9 of an integer, round it to the nearest integer
+		if sdkmath.Abs(exitAmt-sdkmath.Round(exitAmt)) < 1e-9 {
+			exitAmt = (sdkmath.Round(exitAmt))
+		}
```
p0mvn, line 616: Should we truncate in the pool's favor in case of rounding to the nearest int?

The general rule is that value out should be rounded down. Value in should be rounded up

pools/usecase/pools_usecase.go
p0mvn: Might be worth adding a test to cover the overflow

p0mvn (approved):
Nice work! I have some light overflow and rounding concerns but LGTM once addressed

p0mvn (changes requested):
Pressed approved a bit early 🙏 

Would like to come back to review the updates around rounding/overflow before final ack

pools/usecase/pools_usecase.go
deividaspetraitis: Good catch! Updated and included owerflow test. :+1:

pools/usecase/pools_usecase.go
deividaspetraitis: It's mentoned `CalcExitPool`, but additionally I have included a link. This is what you were asking originally or it's something else?

pools/usecase/pools_usecase.go
deividaspetraitis: I was aiming to make original test case`TestCalcExitPool` to pass after after conversion to floats, to ensure there is no errors in the conversion logic and `TestCalcExitPool` functions as it was before making current changes.

There are two cases with rounding up/down, for example:

```
// up
89.99999999999999 - should become to 90
119.99999999999999 - should become to 120
59.99999999999999  - should become to 60

// and

// down
209.958 - should become to 209
279.944 - should become to 279
139.972 - should become to 139
```

Omitting rounding logic above tests would start complain as follows;

```
--- FAIL: TestPoolsUsecaseTestSuite (1.37s)
    --- FAIL: TestPoolsUsecaseTestSuite/TestCalcExitPool (0.01s)
        pools_usecase_test.go:707: 
                Error Trace:    /home/deividas/go/src/github.com/deividaspetraitis/sqs/pools/usecase/pools_usecase_test.go:707
                Error:          Not equal: 
                                expected: "900000bar,1200000baz,600000foo"
                                actual  : "899999bar,1200000baz,600000foo"
```

deividaspetraitis:
Addressed requested changes and updated visuals in original PR description. May I have one more reviev round, please? cc @p0mvn

pools/usecase/pools_usecase.go
p0mvn: I see - I'm happy to keep this as is since this is only needed for display so rounding behavior here is not critical.

Thanks for looking into it!

p0mvn (approved):
Great work!

PR #454, opened 2024-08-13
fix multi-hop /custom-direct-quote bug

p0mvn:
Fixes a logic bug in the multi-hop custom direct quote.

With custom direct quotes, there should always be one route consisting of multiple pools.

The original code has been assuming something else. As a result, the downstream logic would break.

## Testing

- Unit tests updates
- Basic integration tests added

## Summary by CodeRabbit

- **New Features**
	- Enhanced functionality for handling multi-hop custom direct quotes, improving accuracy and reliability.
	- Added constants for referencing specific liquidity pools.

- **Bug Fixes**
	- Resolved issues related to multi-hop custom direct quote processing, ensuring stricter validation and improved error handling.

- **Tests**
	- Introduced new test cases for validating direct quotes in single and multi-hop scenarios, expanding test coverage for the quoting mechanism.


deividaspetraitis (approved):

router/usecase/router_usecase.go
```diff
@@ -511,14 +514,36 @@ func (r *routerUseCaseImpl) GetCustomDirectQuoteMultiPool(ctx context.Context, t
...
+		if len(poolsInRoute) != 1 {
+			return nil, fmt.Errorf("custom direct quote route must have 1 pool, had: %d", len(poolsInRoute))
+		}
+
 		// the amountOut value is the amount out of last the tokenOutDenom
 		result.AmountOut = quote.GetAmountOut()
 
 		// append each pool to the route
-		result.Route = append(result.Route, quote.GetRoute()...)
+		pools = append(pools, poolsInRoute...)
```
deividaspetraitis, line 531: Multi route issue in the tess was resolved by by requesting only "single" route quotes, returning an error is a definitely good practise to indicate client about limits of API.

However I am not following how limiting to a single pool resolves the issue?

deividaspetraitis: Okay, I think I have figured out, let's day in denom is `uosmo` for a quote request. 

Since routes were returned as split-routes, and let's say each route contained a single pool ( for simplicity ), a pool in denom is always going to be `uosmo`, when we expect that inputs from one route to another will be swapped ( [input = output](https://github.com/osmosis-labs/sqs/blob/037279bd8b8378548c6b0e86d4eb2e83e9041b36/router/usecase/route/route.go#L100)).

Actually [pool.CalcSpotPrice](https://github.com/osmosis-labs/sqs/blob/037279bd8b8378548c6b0e86d4eb2e83e9041b36/router/usecase/route/route.go#L64) will complain about such mistake and return an error such as:

```
error(*errors.errorString) *{
        s: "base asset denom (uosmo) is not in pool with (factory/osmo1em6xs...+145 more",}
(dlv) 
```

But we do not handle this error for such scenario.

deividaspetraitis: I have opened proposal PR to be merged into this one to log such errors https://github.com/osmosis-labs/sqs/pull/456, because to handle such errors seems not a trivial task, because we do not want to fail entire quote in case calculating spot price for a single pool fails.

p0mvn: Let's align on route definitions. There is a "multi hop route" and there are "split routes".

Multi hop route is a single route over multiple pools.

Split routes are splits over multiple routes in proportions of the original amount in.

`/custom-direct-quote` only supports "multi-hop routes". Split routes are not supported.

> However I am not following how limiting to a single pool resolves the issue?

What was done in this PR is to ensure that we construct a "multi hop route" from separate quote responses to a single pool route.

> Since routes were returned as split-routes

The routes are returned as a slice of a split route interface but those might not even be split routes under the hood. The naming of the interface seems to be chosen poorly and causes confusion.

Let me know if any follow up thoughts on this 🙏

PR #456, opened 2024-08-13
SQS-267 | Error logging

deividaspetraitis:
Add more logging for error returned by spot price calculation. This error is otherwise hard to handle and logging at this moment seems like one of the best ways to catch that something is going wrong.


deividaspetraitis:
would be nice to be capable to just call `log.Error/etc` ( as with STD  ) instead of pulling dependency everywhere, but at this point I was not able to came up with better implementation.

p0mvn (approved):

PR #457, opened 2024-08-17
fix: consistently return formatted json response with status codes

p0mvn:
This PR is a follow up to "in given out" refactor to consistently return HTTP status codes as json responses.

Without these,  the HTTP API responses lack context about the error details.  

## Summary by CodeRabbit

- **New Features**
	- Improved API documentation for the `/custom-direct-quote` endpoint for better clarity.
- **Bug Fixes**
	- Corrected response status codes for the `/quote` and `/custom-direct-quote` endpoints to ensure proper API behavior.
	- Enhanced error handling for `GetOptimalQuote` and `GetDirectCustomQuote` methods to provide structured JSON responses on errors.
- **Refactor**
	- Simplified error handling in the `GetDirectCustomQuoteRequest` and `GetQuoteRequest` methods to ensure consistent error propagation.

### DIFF
--- a/router/types/get_quote_request_test.go
+++ b/router/types/get_quote_request_test.go
@@ -1,9 +1,7 @@
 package types_test
 
 import (
-	"net/http"
 	"net/http/httptest"
-	"strings"
 	"testing"
 
 	"github.com/labstack/echo/v4"
@@ -21,9 +19,7 @@ func TestGetQuoteRequestUnmarshal(t *testing.T) {
 		name           string
 		queryParams    map[string]string
 		expectedResult *types.GetQuoteRequest
-		expectedError  error
-		expectedStatus int
-		expectedBody   string
+		expectedError  bool
 	}{
 		{
 			name: "valid request with tokenIn and tokenOut",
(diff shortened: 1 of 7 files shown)


router/types/get_quote_request_test.go
```diff
@@ -112,20 +101,15 @@ func TestGetQuoteRequestUnmarshal(t *testing.T) {
 			var result types.GetQuoteRequest
 			err := (&result).UnmarshalHTTPRequest(c)
```
deividaspetraitis, line 102: Omiting http response, status code check makes  sense, since we are testing `UnmarshalHTTPRequest` that returns an error. But why not check `err` returned from `UnmarshalHTTPRequest` against  `expectedError` as it was previously?  Checking only whether it returns and error or not does not allow to check errors that may be returned to the clients and may break compatability.

deividaspetraitis (commented):
Great job, now it looks much more consistent and cleaner now! :+1: I just have one concern regarding `UnmarshalHTTPRequest` test I have left comment for.

router/types/get_quote_request_test.go
p0mvn: Let's track this in a task. Would like to get the bug fix to production

PR #459, opened 2024-08-19
Implement fetching user's active orders query in SQS

deividaspetraitis:
This PR introduces a new `GetActiveOrders` method for fetching active orders by owner from the orderbook contract. 

Additionally refactors response returned by `GetTickUnrealizedCancels` and adds `OrderbookGRPCClientMock`.

### DIFF
--- a/domain/orderbook/plugin/orderbook_tick.go
+++ b/domain/orderbook/plugin/orderbook_tick.go
@@ -1,9 +1,0 @@
-package orderbookplugindomain
-
-import orderbookdomain "github.com/osmosis-labs/sqs/domain/orderbook"
-
-// UnrealizedTickCancels is a struct that represents the response payload from orderbook for unnrealized cancels by tick.
-type UnrealizedTickCancels struct {
-	TickID                 int64                             `json:"tick_id"`
-	UnrealizedCancelsState orderbookdomain.UnrealizedCancels `json:"unrealized_cancels"`
-}
--- a/domain/orderbook/plugin/orderbook_tick.go
+++ b/domain/orderbook/plugin/orderbook_tick.go
@@ -1,0 +1,9 @@
+package orderbookplugindomain
+
+import orderbookdomain "github.com/osmosis-labs/sqs/domain/orderbook"
+
+// UnrealizedTickCancels is a struct that represents the response payload from orderbook for unnrealized cancels by tick.
+type UnrealizedTickCancels struct {
+	TickID                 int64                             `json:"tick_id"`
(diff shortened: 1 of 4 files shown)


domain/orderbook/plugin/orderbook_tick.go
```diff
@@ -0,0 +1,9 @@
+package orderbookplugindomain
+
+import orderbookdomain "github.com/osmosis-labs/sqs/domain/orderbook"
+
+// UnrealizedTickCancels is a struct that represents the response payload from orderbook for unnrealized cancels by tick.
+type UnrealizedTickCancels struct {
+	TickID                 int64                             `json:"tick_id"`
+	UnrealizedCancelsState orderbookdomain.UnrealizedCancels `json:"unrealized_cancels"`
+}
```
p0mvn: What was the motivation for moving it here?

`orderbookplugindomain` is for anything relevant to the fill bot (plugin). The fill bot does not use this type so the original package `orderbookgrpcclientdomain` might be a better fit because it is more general.

p0mvn (approved):
LGTM - nice work - only one nit

domain/orderbook/plugin/orderbook_tick.go
deividaspetraitis: I have refactored because `GetTickUnrealizedCancels` returned unexported data type making imposible to implement for Mock for the client. As a result I made it consistent with other metods as well, instead just by making old data structure exported.

p0mvn: I see - that makes sense but we should still be able to move this back to `orderbookgrpcclientdomain` it sounds like. Could we please do it?

deividaspetraitis: I am still feel confused, all other [OrderBookClient](https://github.com/osmosis-labs/sqs/blob/130a34edb41f978f649cbdc0d7ad5ef7dd01616d/domain/orderbook/grpcclient/orderbook_grpc_client.go#L13) methods uses data structures from `orderbookplugindomain`, and this package has its own gRPC clients, such as [OrderbookCWAPIClient](https://github.com/osmosis-labs/sqs/blob/130a34edb41f978f649cbdc0d7ad5ef7dd01616d/domain/orderbook/plugin/orderbook_grpc_client.go#L8). Addressed request with [130a34e](https://github.com/osmosis-labs/sqs/pull/459/commits/130a34edb41f978f649cbdc0d7ad5ef7dd01616d), merging this PR to unblock us moving forward. :rocket:

PR #460, opened 2024-08-19
refactor: configs

p0mvn:
This PR refactors and improves config management in SQS.

## Changes
- remove unused debug config and flag
- introduce sane defaults for all config values, continuing to allow providing overrides via `config.json` (backward compatible)
- dynamically create environment variable bindings for every config value. For example, `router.enabled` translates to `SQS_ROUTER_ENABLED`
- switch fill bot to environment variable configuration

The precedence is (highest overwrites lowest):
- Environment variable overrides
- Config overrides
- Hardcoded default

Outcomes of discussion [here](https://osmosis-network.slack.com/archives/C079LNKRUCB/p1724090219559519?thread_ts=1724089081.667819&cid=C079LNKRUCB)

## Backwards Compatibility

Fully backwards compatible with the existing config. The old config values would be replaced as overrides.

## Testing

- Tested config override locally
- Tested no config override locally
- Tested override via environment variables locally
- Tested fill bot locally
- Tested copying `config.json` from stage and observing values replaced as intended

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive configuration management system, allowing users to set and override configuration values through files and environment variables.
	- New documentation detailing configuration settings has been added.

- **Improvements**
	- Streamlined configuration files by removing unused settings and simplifying command executions in Docker and Makefile.
	- Enhanced error handling for missing configuration files.
	- Improved logging behavior during application startup for clearer output.

- **Documentation**
	- Expanded instructions on configuration management in `docs/architecture/config.md` and updated `README.md` to redirect users to this new documentation.

- **Bug Fixes**
	- Adjusted logging and configuration handling for improved clarity and usability.
	- Updated test cases to reflect changes in expected outcomes, ensuring accuracy and efficiency in testing.


domain/config.go
```diff
@@ -60,16 +60,126 @@ type Config struct {
 	CORS *CORSConfig `mapstructure:"cors"`
 }
 
+var (
+	DefaultConfig = Config{
```
p0mvn, line 67: Note to reviewer: This is the new default configuration. Most values should be ported from the old `config.json`.

They are overridden by:
- env vars
- `config.json`

ingest/usecase/plugins/orderbookfiller/create_copy_config.sh
```diff
@@ -1,19 +1,5 @@
...
-# Use jq to modify the JSON and create a new file
-jq '. 
-    | .["grpc-tendermint-rpc-endpoint"] = "http://osmosis:26657" 
-    | .["grpc-gateway-endpoint"] = "osmosis:9090" 
-    | .otel.environment = "sqs-fill-bot" 
-    | .["grpc-ingester"].plugins[] |= if .name == "orderbook" then .enabled = true else . end' \
-    "$INPUT_FILE" > "$OUTPUT_FILE"
-
-echo "Modified configuration saved to $OUTPUT_FILE"
-
```
p0mvn, line 16: Note to reviewer: fill bot switched to environment variables instead of file config

domain/config.go
```diff
@@ -121,6 +245,48 @@ func viperDecodeHookFunc() mapstructure.DecodeHookFunc {
...
+			bindEnvRecursive(value, envName+".")
+		} else {
+			// Bind the environment variable
+			if err := viper.BindEnv(envName); err != nil {
+				panic(err)
+			}
+		}
+	}
+}
+
```
p0mvn, line 289: Note to reviewer: this logic handles dynamically creating environment variable bindings to to allow overriding any config value from the default config

domain/config.go
```diff
@@ -60,16 +61,139 @@ type Config struct {
...
 func UnmarshalConfig() (*Config, error) {
-	var config Config
+	config := DefaultConfig
+
+	viper.SetEnvPrefix(envPrefix)
+	viper.AutomaticEnv()
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+
+	// Set up environment variable mappings using reflection
+	bindEnvRecursive(reflect.ValueOf(&config), "")
```
p0mvn, line 189: Note to reviewer: this logic handles dynamically creating environment variable bindings to to allow overriding any config value from the default config

docs/architecture/config.md
cryptomatictrader, line 1: Very nice explanation! 👍

cryptomatictrader (approved):
Nice work! Shall we print out all SQS configurations during SQS starts up given that a single config can come from either default, config.json or env var?

Have to closely watch staging and production in our next deployment as always.

Thank you!

domain/config.go
pablin-10: from my perspective, the priority should be:
cli parameters > env vars > config file > in-code defaults

niccoloraspa: If we have both CLI parameters and environment variables, it can become a bit confusing.

In my opinion, environment variables (env vars) should have the highest priority. For example, Terraform prioritizes environment variables, giving them higher precedence, while CLI parameters have the lowest precedence (even lower than the config). However, this is just a matter of preference, and it can become quite confusing.

I think we should avoid this altogether and leave it as it is, without allowing the config to be changed via the CLI.

p0mvn: That sounds good. Let's go with the current approach (no CLI parameters for now), and we can iterate in future PRs as needed.

p0mvn:
> Nice work! Shall we print out all SQS configurations during SQS starts up given that a single config can come from either default, config.json or env var?
> 
> Have to closely watch staging and production in our next deployment as always.
> 
> Thank you!

Thanks. We have a `/config` endpoint to check so logs might be an overkill but good suggestion

PR #468, opened 2024-08-23
fix: active orders overflow bug

p0mvn:
Discovered overflow when computing the price scaling factor. Instead, we can get it from cache.

```
curl -X 'GET'     'http://localhost:9092/passthrough/active-orders?userOsmoAddress=osmo1q8709l2656zjtg567xnrxjr6j35a2pvwhxxms2'     -H 'accept: application/json' | jq .
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced order creation process with improved error handling for normalization factors.
	- Updated method for calculating normalized prices to increase robustness.

- **Bug Fixes**
	- Resolved issues related to the previous normalization factor calculation, improving overall order processing logic.


deividaspetraitis (approved):
Looks good to me, I have also verified that change does not effect output by runnig a diff on after/before responses.

PR #471, opened 2024-08-23
active order fault tolerance via "best-effort" boolean return 

p0mvn:
Add fault tolerance to the active order query.

This is so that if at least one order is skipped due to an error, the client is notified that the result is " best effort."

## Summary by CodeRabbit

- **New Features**
	- Enhanced `GetActiveOrders` method to return an additional boolean indicating whether the order retrieval was a best-effort attempt.
	- Updated response structure to include a new field, `IsBestEffort`, providing more context about the order retrieval process.

- **Bug Fixes**
	- Improved error handling and status reporting in order processing methods to enhance robustness.

- **Documentation**
	- Updated method signatures and response structures to reflect the new boolean return values and fields.


deividaspetraitis (approved):
Merged with `v25.x` and resolved conflicts, tests are passing, merging in.

PR #480, opened 2024-08-27
fix: no routes for this trade

cryptomatictrader:
This commit addresses the issue observed on FE showing the error "No routes for this trade". The changes allow the route search loop continue even the denom data is unavailable for the denom of pool in the route being searched. 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved logging for candidate route finding, providing better debugging information when no pools are found.
- **Tests**
	- Updated test case to reflect changes in the expected number of unsupported tokens, enhancing test accuracy and relevance.


p0mvn (approved):
Trusting that we confirmed this change with a manual integration test run per-discussions.

Nice work on debugging and thank you!

PR #492, opened 2024-08-28
SQS-375 | Unit test createFormattedLimitOrder

deividaspetraitis:
This PR introduces unit tests for Orderbook usecase `createFormattedLimitOrder` method. The aim of the tests are to cover edge cases as well as successful scenario. As a side effect pull request introduces `OrderbookRepositoryMock`.

```
❯ go test ./orderbook/usecase -run "TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder" -v
=== RUN   TestOrderbookUsecaseTestSuite
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/tick_not_found
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_quantity
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_placed_quantity
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_getting_spot_price_scaling_factor
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_bid_effective_total_amount_swapped
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_bid_unrealized_cancels
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_etas
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_converting_tick_to_price
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_placed_at
=== RUN   TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/successful_order_processing
--- PASS: TestOrderbookUsecaseTestSuite (0.03s)
    --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder (0.03s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/tick_not_found (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_quantity (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_placed_quantity (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_getting_spot_price_scaling_factor (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_bid_effective_total_amount_swapped (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_bid_unrealized_cancels (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_etas (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_converting_tick_to_price (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/error_parsing_placed_at (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestCreateFormattedLimitOrder/successful_order_processing (0.00s)
PASS
ok      github.com/osmosis-labs/sqs/orderbook/usecase   0.090s
```

## Summary by CodeRabbit

- **New Features**
	- Introduced a mock implementation for the OrderBookRepository to enhance testing capabilities.
	- Added a public interface for creating formatted limit orders in the order book domain.
	- Implemented a comprehensive test suite for the OrderbookUsecase functionality, covering various scenarios for order creation.
	- Defined custom error types for improved error handling and reporting within the order book system.
	- Enhanced the OrderBookClient interface with new methods for fetching unrealized cancels and ticks in chunks.
	- Updated the LimitOrder struct to use higher precision decimal types for key fields.
	- Added a new endpoint to retrieve active orders associated with a specified Osmo wallet address.

- **Bug Fixes**
	- Added validation for `placedQuantity` in order creation to prevent invalid orders.
	- Adjusted return types for better performance and type handling in order book use case functions.

### DIFF
--- a/orderbook/usecase/orderbook_usecase.go
+++ b/orderbook/usecase/orderbook_usecase.go
@@ -3,7 +3,6 @@ package orderbookusecase
 import (
 	"context"
 	"fmt"
-	"math"
 	"strconv"
 	"time"
 
@@ -276,7 +275,7 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 	tickState := tickForOrder.TickState
 	unrealizedCancels := tickForOrder.UnrealizedCancels
 
-	quantity, err := strconv.ParseInt(order.Quantity, 10, 64)
+	quantity, err := osmomath.NewDecFromStr(order.Quantity)
 	if err != nil {
 		return orderbookdomain.LimitOrder{}, types.ParsingQuantityError{
 			Quantity: order.Quantity,
@@ -284,10 +283,7 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 		}
 	}
 
-	// Convert quantity to decimal for the calculations
-	quantityDec := osmomath.NewDec(quantity)
-
-	placedQuantity, err := strconv.ParseInt(order.PlacedQuantity, 10, 64)
+	placedQuantity, err := osmomath.NewDecFromStr(order.PlacedQuantity)
 	if err != nil {
 		return orderbookdomain.LimitOrder{}, types.ParsingPlacedQuantityError{
 			PlacedQuantity: order.PlacedQuantity,
@@ -295,7 +291,7 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 		}
 	}
 
-	if placedQuantity == 0 || placedQuantity < 0 {
+	if zero := osmomath.NewDec(0); placedQuantity.Equal(zero) || placedQuantity.LT(zero) {
 		return orderbookdomain.LimitOrder{}, types.InvalidPlacedQuantityError{PlacedQuantity: placedQuantity}
 	}
 
@@ -308,7 +304,7 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 	}
 
 	// Calculate percent claimed
-	percentClaimed := placedQuantityDec.Sub(quantityDec).Quo(placedQuantityDec)
+	percentClaimed := placedQuantityDec.Sub(quantity).Quo(placedQuantityDec)
 
 	// Calculate normalization factor for price
 	normalizationFactor, err := o.tokensUsecease.GetSpotPriceScalingFactorByDenom(baseAsset.Symbol, quoteAsset.Symbol)
@@ -367,16 +363,16 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 
 	tickTotalEtas := tickEtas + tickUnrealizedCancelled
 
-	totalFilled := int64(math.Max(
-		float64(tickTotalEtas-(etas-(placedQuantity-quantity))),
-		0,
-	))
+	totalFilled := osmomath.MaxDec(
+		osmomath.NewDec(tickTotalEtas).Sub(osmomath.NewDec(etas).Sub(placedQuantity.Sub(quantity))),
+		osmomath.ZeroDec(),
+	)
 
 	// Calculate percent filled using
-	percentFilled, err := osmomath.NewDecFromStr(strconv.FormatFloat(math.Min(float64(totalFilled)/float64(placedQuantity), 1), 'f', -1, 64))
-	if err != nil {
-		return orderbookdomain.LimitOrder{}, types.CalculatingPercentFilledError{Err: err}
-	}
+	percentFilled := osmomath.MinDec(
+		totalFilled.Quo(placedQuantity),
+		osmomath.OneDec(),
+	)
 
 	// Determine order status based on percent filled
 	status, err := order.Status(percentFilled.MustFloat64())
@@ -421,13 +417,13 @@ func (o *OrderbookUseCaseImpl) createFormattedLimitOrder(
 		Etas:             order.Etas,
 		ClaimBounty:      order.ClaimBounty,
 		PlacedQuantity:   placedQuantity,
-		PercentClaimed:   percentClaimed.String(),
+		PercentClaimed:   percentClaimed,
 		TotalFilled:      totalFilled,
-		PercentFilled:    percentFilled.String(),
+		PercentFilled:    percentFilled,
 		OrderbookAddress: orderbookAddress,
-		Price:            normalizedPrice.String(),
+		Price:            normalizedPrice,
 		Status:           status,
-		Output:           output.String(),
+		Output:           output,
 		QuoteAsset:       quoteAsset,
 		BaseAsset:        baseAsset,
 		PlacedAt:         placedAt,
--- a/orderbook/usecase/orderbook_usecase.go
+++ b/orderbook/usecase/orderbook_usecase.go
@@ -15,6 +15,7 @@ import (
 	orderbookgrpcclientdomain "github.com/osmosis-labs/sqs/domain/orderbook/grpcclient"
 	"github.com/osmosis-labs/sqs/log"
 	"github.com/osmosis-labs/sqs/orderbook/telemetry"
+	"github.com/osmosis-labs/sqs/orderbook/types"
 	"github.com/osmosis-labs/sqs/sqsdomain"
 	"go.uber.org/zap"
 
@@ -262,16 +263,21 @@ func (o *orderbookUseCaseImpl) createFormattedLimitOrder(
 	tickForOrder, ok := o.orderbookRepository.GetTickByID(poolID, order.TickId)
 	if !ok {
 		telemetry.GetTickByIDNotFoundCounter.Inc()
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("tick not found %s, %d", orderbookAddress, order.TickId)
+		return orderbookdomain.LimitOrder{}, types.TickForOrderbookNotFoundError{
+			OrderbookAddress: orderbookAddress,
+			TickID:           order.TickId,
+		}
 	}
 
 	tickState := tickForOrder.TickState
 	unrealizedCancels := tickForOrder.UnrealizedCancels
 
-	// Parse quantity as int64
 	quantity, err := strconv.ParseInt(order.Quantity, 10, 64)
 	if err != nil {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("error parsing quantity: %w", err)
+		return orderbookdomain.LimitOrder{}, types.ParsingQuantityError{
+			Quantity: order.Quantity,
+			Err:      err,
+		}
 	}
 
 	// Convert quantity to decimal for the calculations
@@ -279,16 +285,22 @@ func (o *orderbookUseCaseImpl) createFormattedLimitOrder(
 
 	placedQuantity, err := strconv.ParseInt(order.PlacedQuantity, 10, 64)
 	if err != nil {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("error parsing placed quantity: %w", err)
+		return orderbookdomain.LimitOrder{}, types.ParsingPlacedQuantityError{
+			PlacedQuantity: order.PlacedQuantity,
+			Err:            err,
+		}
 	}
 
 	if placedQuantity == 0 || placedQuantity < 0 {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("placed quantity is 0 or negative")
+		return orderbookdomain.LimitOrder{}, types.InvalidPlacedQuantityError{PlacedQuantity: placedQuantity}
 	}
 
 	placedQuantityDec, err := osmomath.NewDecFromStr(order.PlacedQuantity)
 	if err != nil {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("error parsing placed quantity: %w", err)
+		return orderbookdomain.LimitOrder{}, types.ParsingPlacedQuantityError{
+			PlacedQuantity: order.PlacedQuantity,
+			Err:            err,
+		}
 	}
 
 	// Calculate percent claimed
@@ -297,7 +309,11 @@ func (o *orderbookUseCaseImpl) createFormattedLimitOrder(
 	// Calculate normalization factor for price
 	normalizationFactor, err := o.tokensUsecease.GetSpotPriceScalingFactorByDenom(baseAsset.Symbol, quoteAsset.Symbol)
 	if err != nil {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("error getting spot price scaling factor: %w", err)
+		return orderbookdomain.LimitOrder{}, types.GettingSpotPriceScalingFactorError{
+			BaseDenom:  baseAsset.Symbol,
+			QuoteDenom: quoteAsset.Symbol,
+			Err:        err,
+		}
 	}
 
 	// Determine tick values and unrealized cancels based on order direction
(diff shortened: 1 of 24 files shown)


orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+			usecase := orderbookusecase.New(
+				&orderbookRepo,
+				nil,
+				nil,
+				&tokensUsecase,
+				nil,
+			)
+
+			// Call the method under test
+			result, err := usecase.CreateFormattedLimitOrder(1, tc.order, tc.quoteAsset, tc.baseAsset, "someOrderbookAddress")
```
p0mvn: Let's define a constant `defaultPoolID`. Also, does it make sense to have at least one test with a different poolID and `orderbokAdddress` to ensure that the test does not break with default values.

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+		setupMocks    func(orderbookRepo *mocks.OrderbookRepositoryMock, tokensUsecase *mocks.TokensUsecaseMock)
+		expectedError string
+		expectedOrder orderbookdomain.LimitOrder
+	}{
+		{
+			name: "tick not found",
+			order: orderbookdomain.Order{
+				TickId: 99, // Non-existent tick ID
+			},
+			expectedError: "tick not found",
```
p0mvn: Instead of hardcoding error string, consider referencing an actual error returned from core codebase that we can define as a var and export via `export_test.go`

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+						},
+					}, true
+				}
+			},
+		},
+		{
+			name: "error getting spot price scaling factor",
+			order: orderbookdomain.Order{
+				Quantity:       "1000",
+				PlacedQuantity: "1500",
```
p0mvn: Consider reducing code duplication and hardcoding by defining constants

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+				Quantity: "invalid", // Invalid quantity
+			},
+			expectedError: "error parsing quantity",
+			setupMocks: func(orderbookRepo *mocks.OrderbookRepositoryMock, tokensUsecase *mocks.TokensUsecaseMock) {
+				orderbookRepo.GetTickByIDFunc = func(poolID uint64, tickID int64) (orderbookdomain.OrderbookTick, bool) {
+					return orderbookdomain.OrderbookTick{
+						Tick: &cosmwasmpool.OrderbookTick{
+							TickId: tickID,
+						},
+					}, true
```
p0mvn: The mocked `GetTickByIDFunc` seems repeating from the previous test. Would it be possible to define it on top? This would reduce code duplication, increasing maintainability and readability

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+							TickId: tickID,
+						},
+					}, false
+				}
+			},
+		},
+		{
+			name: "error parsing quantity",
+			order: orderbookdomain.Order{
+				Quantity: "invalid", // Invalid quantity
```
p0mvn, line 643: Could we also add a test where quantity is a large int that overflows and handled gracefully?

Thinking more about this, I think this is very likely to overflow and panic in prod. Could we instead do string to Dec conversion via `NewDecFromStr` while also adding a relevant test?

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+					return orderbookdomain.OrderbookTick{
+						Tick: &cosmwasmpool.OrderbookTick{
+							TickId: tickID,
+						},
+					}, true
+				}
+			},
+		},
+		{
+			name: "error parsing placed quantity",
```
p0mvn, line 666: I think that we need to change the way we parse quantity for placed quantity as well `NewDecFromStr`

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+			order: orderbookdomain.Order{
+				Quantity:       "1000",
+				PlacedQuantity: "invalid", // Invalid placed quantity
+			},
+			expectedError: "error parsing placed quantity",
```
p0mvn: We should have a test where placed quantity is zero. Currently, the system will panic due to division by zero, and we need to add a guarding check against the panic

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,314 @@
...
+			order: orderbookdomain.Order{
+				TickId:         1,
+				OrderId:        1,
+				OrderDirection: "bid",
+				Owner:          "owner1",
+				Quantity:       "1000",
+				PlacedQuantity: "1500",
+				Etas:           "500",
+				ClaimBounty:    "10",
+				PlacedAt:       "1634764800000",
```
p0mvn: Let's review and revalidate all numerical values against overflows

p0mvn (commented):
Nice work. Left some comments - please take a look.

Mostly, recommendations around overflows and reducing code duplication

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: PoolID is passed because it is requires as argument, actual function that uses this argument is mocked, thus it can be any number value. To avoid use of constant I have moved it to `tc.poolID` that defauts to `0` at this moment for all test cases.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: That is a valid point, my initial attempt implementing that failed, since for this method client(s) specific errors are irrelevant, and added complexity by defining errors and introduced additional code outweights the benefts, thus I would leave as it is at least for now. Let me know if you still think we should avoid doing that.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: The main reason I am not reusing contants here was to make values local to the test case to avoid adding cognitive load. Let me know if that makes sense.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Good point! I have just did that.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Good catch! Added overflow checks. :rocket:

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Good catch! Added test to cover this case. :+1:

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: We do [return](https://github.com/osmosis-labs/sqs/blob/e4ce75541f31c4de7baf6da3e8b95fae6744f03e/orderbook/usecase/orderbook_usecase.go#L383) quantity as `int64` for the clients, if for example `quantity` overflows we will catch it with `strconv.ParseInt`. At this point I am trying to have exactly same precision logic as FE had. If we want to have some adjusments regarding that, maybe we can create a separate task to track that?

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Good point, implemented! :+1:

orderbook/usecase/orderbook_usecase_test.go
p0mvn: I still think it is worth it doing something like [this](https://github.com/osmosis-labs/sqs/blob/f1740c12a07ac7a2c9df284d45b1323224a755ef/domain/errors.go#L318) so that if an error message is ever updated, we only need to update it once and don't worry about several hardcoded tests

orderbook/usecase/orderbook_usecase_test.go
p0mvn: My point is that it will frequently overflow in `strconv.ParseInt`, making the system unusable.

Consider 1000 DAI which is 1000 * 10**18 in base units.

```
>>> 2**63 - 1 - 1000 * 10**18
-990776627963145224193
```

Just $1000 of DAI overflows here so we should use the appropriate chain types to prevent that.

deividaspetraitis: Ah now I see you point, thanks for elaboration! :pray: The idea is that currently we support only stables ( `USDC`, `USDT` ) as a quotes each of which has `6` decimal places,  meaning that paid amount should be larger than `1*10**12` in order to overflow:

```
>>> 2**63 - 1 - 1*10**13 * 10**6
-776627963145224193
```

In order to support non stables as a quotes would require some adjustments on FE side to, meaning we can be reasurred that somebody from product will not enable support of more asset types without anyone noticing from engineering team ( including us ) in advance.

Having this information, do you think we still should go and implement this change within scope of this task?

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Thanks, I have implemented as per your suggestion [here](https://github.com/osmosis-labs/sqs/pull/492/files#diff-a585202576c1e374948b8b760a1fe96534e5a35a29ccb3e846e5a7548c714f0a). Would that work? :pray:

orderbook/usecase/orderbook_usecase_test.go
p0mvn: I'm not sure I understand how stables as quotes excludes the possibility of overflow.

An order can be placed in both base and quote denominations so amounts are not limited to only the listed stables precision. 

Additionally, there are 18 decimal stables like DAI which we can trivially add support for.

I would vote to do it right now so that we don't need to worry about it. I also have low confidence in our cross-team communication on breaking changes so being extra diligent around such edge cases would pay-off down the road, helping to avoid incidents.

orderbook/usecase/orderbook_usecase_test.go
p0mvn: yes, nice job!

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Thanks! I have implemented your suggested improvement [here](https://github.com/osmosis-labs/sqs/pull/492/files#diff-ae675fb65861261e203f8bf1a1aacfd68815f3cb763d28db817797e9a88904b3R278) and have adjusted tests to check cases for possible `osmomath.Dec` overflows. Would that work?

orderbook/usecase/orderbook_usecase.go
```diff
@@ -262,94 +266,124 @@ func (o *orderbookUseCaseImpl) createFormattedLimitOrder(
...
+	placedQuantity, err := osmomath.NewDecFromStr(order.PlacedQuantity)
 	if err != nil {
-		return orderbookdomain.LimitOrder{}, fmt.Errorf("error parsing placed quantity: %w", err)
+		return orderbookdomain.LimitOrder{}, types.ParsingPlacedQuantityError{
+			PlacedQuantity: order.PlacedQuantity,
+			Err:            err,
+		}
+	}
+
+	if zero := osmomath.NewDec(0); placedQuantity.Equal(zero) || placedQuantity.LT(zero) {
```
p0mvn: nit: we could define `zero := osmomath.NewDec(0)` in the global space so that it does not have to be re-initilized for every client call

orderbook/usecase/orderbook_usecase.go
```diff
@@ -262,94 +266,124 @@ func (o *orderbookUseCaseImpl) createFormattedLimitOrder(
...
+		}
 	}
 
 	// Determine tick values and unrealized cancels based on order direction
 	var tickEtas, tickUnrealizedCancelled int64
 	if order.OrderDirection == "bid" {
 		tickEtas, err = strconv.ParseInt(tickState.BidValues.EffectiveTotalAmountSwapped, 10, 64)
 		if err != nil {
-			return orderbookdomain.LimitOrder{}, fmt.Errorf("error parsing bid effective total amount swapped: %w", err)
+			return orderbookdomain.LimitOrder{}, types.ParsingTickValuesError{
```
p0mvn, line 332: I think we might also change the parsing for tick ETAs.

Synched with Connor on definition:
> There's two places where we record an effective total amount swapped value. One is in the tick (for each direction) and one is in the order itself. When you place an order the current tick's ETAS (effective total amount swapped) value is recorded under the order's etas field. As swaps are performed on the tick we update the tick's ETAS value (for the given direction). The tick's ETAS value is then compared against the order's current etas field to determine how much of the order has been filled. If the order is partially filled we update the etas field to be the current tick's ETAS, if the order is fully filled it's simply removed.

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -0,0 +1,542 @@
...
+			usecase := orderbookusecase.New(
+				&orderbookRepo,
+				nil,
+				nil,
+				&tokensUsecase,
+				nil,
+			)
+
+			// Call the method under test
+			result, err := usecase.CreateFormattedLimitOrder(tc.poolID, tc.order, tc.quoteAsset, tc.baseAsset, "someOrderbookAddress")
```
p0mvn: nit: would be good to have a test case with the address varied for full coverage of the input set

p0mvn (commented):
One blocking comment remaining:
- The concern around ETAs overflow

orderbook/usecase/orderbook_usecase.go
deividaspetraitis: Addressed! :+1:

orderbook/usecase/orderbook_usecase.go
deividaspetraitis: Oh, that's a really good catch, thanks! Updated as per your suggestion. :pray:

p0mvn (approved):
LGTM

Could we address this one dangling comment pre-merge?
https://github.com/osmosis-labs/sqs/pull/492#discussion_r1742422670

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Addressed with [4e1e945](https://github.com/osmosis-labs/sqs/pull/492/commits/4e1e9457636004e1f5bf701246a10526abad37a4) :+1:

PR #495, opened 2024-08-30
SQS-378 | GetActiveOrders

deividaspetraitis:
This PR introduces unit tests for Orderbook usecase `GetActiveOrders` method. The aim of the tests are to cover edge cases as well as successful scenario.

```
❯ go test ./orderbook/usecase -run "TestOrderbookUsecaseTestSuite/TestGetActiveOrders" -v
=== RUN   TestOrderbookUsecaseTestSuite
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders/failed_to_get_all_canonical_orderbook_pool_IDs
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders/context_is_done_before_processing_all_orderbooks
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders/processOrderBookActiveOrders_returns_an_error
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders/isBestEffort_set_to_true_when_one_orderbook_is_processed_with_best_effort
=== RUN   TestOrderbookUsecaseTestSuite/TestGetActiveOrders/successful_retrieval_of_active_orders
--- PASS: TestOrderbookUsecaseTestSuite (0.03s)
    --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders (0.03s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders/failed_to_get_all_canonical_orderbook_pool_IDs (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders/context_is_done_before_processing_all_orderbooks (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders/processOrderBookActiveOrders_returns_an_error (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders/isBestEffort_set_to_true_when_one_orderbook_is_processed_with_best_effort (0.00s)
        --- PASS: TestOrderbookUsecaseTestSuite/TestGetActiveOrders/successful_retrieval_of_active_orders (0.00s)
PASS
ok      github.com/osmosis-labs/sqs/orderbook/usecase   (cached)

```

### DIFF
--- a/orderbook/usecase/orderbook_usecase.go
+++ b/orderbook/usecase/orderbook_usecase.go
@@ -21,7 +21,7 @@ import (
 	clmath "github.com/osmosis-labs/osmosis/v25/x/concentrated-liquidity/math"
 )
 
-type orderbookUseCaseImpl struct {
+type OrderbookUseCaseImpl struct {
 	orderbookRepository orderbookdomain.OrderBookRepository
 	orderBookClient     orderbookgrpcclientdomain.OrderBookClient
 	poolsUsecease       mvc.PoolsUsecase
@@ -29,7 +29,7 @@ type orderbookUseCaseImpl struct {
 	logger              log.Logger
 }
 
-var _ mvc.OrderBookUsecase = &orderbookUseCaseImpl{}
+var _ mvc.OrderBookUsecase = &OrderbookUseCaseImpl{}
 
 const (
 	// Max number of ticks to query at a time
@@ -45,8 +45,8 @@ func New(
 	poolsUsecease mvc.PoolsUsecase,
 	tokensUsecease mvc.TokensUsecase,
 	logger log.Logger,
-) *orderbookUseCaseImpl {
-	return &orderbookUseCaseImpl{
+) *OrderbookUseCaseImpl {
+	return &OrderbookUseCaseImpl{
 		orderbookRepository: orderbookRepository,
 		orderBookClient:     orderBookClient,
 		poolsUsecease:       poolsUsecease,
@@ -56,12 +56,16 @@ func New(
 }
 
 // GetTicks implements mvc.OrderBookUsecase.
-func (o *orderbookUseCaseImpl) GetAllTicks(poolID uint64) (map[int64]orderbookdomain.OrderbookTick, bool) {
+func (o *OrderbookUseCaseImpl) GetAllTicks(poolID uint64) (map[int64]orderbookdomain.OrderbookTick, bool) {
 	return o.orderbookRepository.GetAllTicks(poolID)
 }
 
 // StoreTicks implements mvc.OrderBookUsecase.
-func (o *orderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.PoolI) error {
+func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.PoolI) error {
+	if pool == nil {
+		return fmt.Errorf("pool is nil when processing order book")
+	}
+
 	cosmWasmPoolModel := pool.GetSQSPoolModel().CosmWasmPoolModel
 	if cosmWasmPoolModel == nil {
 		return fmt.Errorf("cw pool model is nil when processing order book")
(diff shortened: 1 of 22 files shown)


orderbook/usecase/orderbook_usecase.go
```diff
@@ -46,13 +48,17 @@ func New(
...
-	return &OrderbookUseCaseImpl{
+	res := OrderbookUseCaseImpl{
 		orderbookRepository: orderbookRepository,
 		orderBookClient:     orderBookClient,
 		poolsUsecease:       poolsUsecease,
 		tokensUsecease:      tokensUsecease,
 		logger:              logger,
 	}
+
+	res.processOrderBookActiveOrdersFunc = res.processOrderBookActiveOrders
```
deividaspetraitis: Ideally we would want to have it as a dependency. Any suggestions?

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -221,6 +223,184 @@ func (s *OrderbookUsecaseTestSuite) TestProcessPool() {
...
+			if tc.setupMocks != nil {
+				tc.setupMocks(usecase, &orderbookRepository, &client, &poolsUsecase, &tokensusecase)
+			}
+
+			ctx := tc.setupContext()
+
+			// Call the method under test
+			// We are not interested in the orders returned, it's tested
+			// in the TestCreateFormattedLimitOrder.
+			_, isBestEffort, err := usecase.GetActiveOrders(ctx, "address")
```
p0mvn: Could we add one test case where the address is different than "address"`?

Per principles of black box testing, we should aim to vary all combinations of inputs. Otherwise, if there is a conditional logic on the specific value of address, our test would never catch it

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -221,6 +223,184 @@ func (s *OrderbookUsecaseTestSuite) TestProcessPool() {
...
+
+			// Assert the results
+			if tc.expectedError != nil {
+				s.Assert().Error(err)
+				if errors.Is(err, tc.expectedError) {
+					s.Assert().ErrorIs(err, tc.expectedError)
+				} else {
+					s.Assert().ErrorAs(err, tc.expectedError)
+				}
```
p0mvn: This looks like a good possible test helper to move over into the `routertesting` package that we reimport in most tests

orderbook/usecase/orderbook_usecase_test.go
p0mvn: Can we also assert the limit order return as opposed to ignoring it?

At the very least we could validate that the number of orders is as expected and IDs match

orderbook/usecase/orderbook_usecase_test.go
```diff
@@ -221,6 +223,184 @@ func (s *OrderbookUsecaseTestSuite) TestProcessPool() {
...
+				// Set is best effort to true
+				orderbookrepository.GetTickByIDFunc = func(poolID uint64, tickID int64) (orderbookdomain.OrderbookTick, bool) {
+					return orderbookdomain.OrderbookTick{}, false
+				}
+			},
+			expectedError:        nil,
+			expectedIsBestEffort: true,
+		},
+		{
+			name: "successful retrieval of active orders",
```
p0mvn: I think that we would benefit from more cases:
1. 3 orders returned. 1 from order book A, 2 from orderbook B -> 3 orders are returned as intended
2. 2 orders returned. 1 from order book A, 1 from order book B. Order book B is not canonical -> only 1 order is returned
3. 2 orders returned. 1 is not from the input address -> 1 order returned

This would require defining more orders.

What we can do is create 1 `defaultOrder` at the top of the test. Then, use a decorator pattern to change them as needed for readability. For example:


```go
withOrderID := func(o Order, poolID uint64) Order {
   o.PoolID = poolID
   return o
}
```

p0mvn (commented):
Nice work! Left a couple of recommendations

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: I have implemented as per your suggestion! Just one question, by `canonical` you meant it's not valid or something else? For the rest case I was not able to implement that because there is no way to check does the order belongs to the input address since order does contain address field.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Implemented to assert actual orders with expected ones. :+1:

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Thanks for suggestion, moved to `routertesting`.

orderbook/usecase/orderbook_usecase_test.go
deividaspetraitis: Actual address is never used in underlying mocking logic, but I have moved address to testcase instance variable and assigned unique address for each test case. :+1:

orderbook/usecase/orderbook_usecase_test.go
p0mvn: > by canonical you meant it's not valid or something else?

Sorry for the confusion - I meant that there is no order in that order book for a given user

p0mvn (approved):
Nice work!

PR #496, opened 2024-09-02
chore: update sqsdomain to tagged v26 osmosis node

PaddyMc:
### Description

update sqs domain to tagged v26

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,8 +14,8 @@ require (
 	github.com/labstack/echo/v4 v4.12.0
 	github.com/osmosis-labs/osmosis/osmomath v0.0.13
 	github.com/osmosis-labs/osmosis/osmoutils v0.0.13
-	github.com/osmosis-labs/osmosis/v26 v26.0.0-20240827120025-dd5be6c09d02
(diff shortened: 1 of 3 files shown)


deividaspetraitis (approved):

go.mod
```diff
@@ -281,8 +281,6 @@ replace (
 	github.com/osmosis-labs/osmosis/x/epochs => github.com/osmosis-labs/osmosis/x/epochs v0.0.5-0.20240825083448-87db4447a1ff
 	github.com/osmosis-labs/osmosis/x/ibc-hooks => github.com/osmosis-labs/osmosis/x/ibc-hooks v0.0.0-20240825083448-87db4447a1ff
 
-	github.com/osmosis-labs/sqs/sqsdomain => ./sqsdomain
```
deividaspetraitis, line 284: Wasn't this one was intended to point to a local copy instead of remote for development so we're on `v26` branch local version, path should be correct anyways?

PaddyMc: I added this to get sqs to compile correctly, I don't think we should use local replaces

PR #510, opened 2024-09-11
feat: uni-directional trading pair taker fee support

cryptomatictrader:
This commit builds on the work done in [PR 8344](https://github.com/osmosis-labs/osmosis/pull/8344).

Now, the taker fee can be set independently for each trade direction. For example, in a trading pair (A, B), the taker fee for A → B can be P, while the fee for B → A can be Q. With this change, there’s no need to lexicographically sort the pair denominations before storing or retrieving taker fee data. Instead, the taker fee map will now contain two separate entries: one for the pair (A, B) with value P, and another for the pair (B, A) with value Q.

Additionally, adjustments have been made to the node-side logic, where the ingester prepares data for SQS. For more details, refer to [PR 8690](https://github.com/osmosis-labs/osmosis/pull/8690).

Tested in v26 with the following scenario:

1. Create v26 in-place testnet node
2. Start v26 node
3. Start v26 sqs
4. Run `make sqs-update-mainnet-state` in sqs project to capture updated taker fee data for step 5.
5. Run 'make test-unit` in sqs


mattverse (approved):
LGTM

PR #518, opened 2024-09-20
SQS-412 | Active Orders Query: SSE

deividaspetraitis:
This PR implements SSE for Active Orders Query endpoint.

Implementation extends existing functionality by looking for `sse` URL param for existing endpoint, for example: `/passthrough/active-orders?userOsmoAddress=osmo1jgz4xmaw9yk9pjxd4h8c2zs0r0vmgyn88s8t6l&sse=1` When there is no `sse` passed, endpoint will return a list of orders as per its original functionlity.

## Release Notes

- **New Features**
  - Introduced Server-Sent Events (SSE) functionality for active orders, allowing real-time updates.
  - Added a new method for streaming active orders, enhancing dynamic interaction with order data.

- **Improvements**
  - Updated logging capabilities within the PassthroughHandler for better error tracking.
  - Streamlined request handling for active orders, improving error management.

- **Bug Fixes**
  - Enhanced error handling during the retrieval of active orders.

These changes improve the overall user experience by providing real-time data and more robust logging.

### DIFF
--- a/delivery/http/event.go
+++ b/delivery/http/event.go
@@ -1,0 +1,98 @@
+package http
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/labstack/echo/v4"
+)
+
+// Event represents Server-Sent Event.
+// SSE explanation: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format
+type Event struct {
+	// ID is used to set the EventSource object's last event ID value.
+	ID []byte
+	// Data field is for the message. When the EventSource receives multiple consecutive lines
+	// that begin with data:, it concatenates them, inserting a newline character between each one.
+	// Trailing newlines are removed.
+	Data []byte
+	// Event is a string identifying the type of event described. If this is specified, an event
+	// will be dispatched on the browser to the listener for the specified event name; the website
+	// source code should use addEventListener() to listen for named events. The onmessage handler
+	// is called if no event name is specified for a message.
+	Event []byte
+	// Retry is the reconnection time. If the connection to the server is lost, the browser will
+	// wait for the specified time before attempting to reconnect. This must be an integer, specifying
+	// the reconnection time in milliseconds. If a non-integer value is specified, the field is ignored.
+	Retry []byte
+	// Comment line can be used to prevent connections from timing out; a server can send a comment
+	// periodically to keep the connection alive.
+	Comment []byte
+}
+
+// MarshalTo marshals Event to given Writer
+func (ev *Event) MarshalTo(w io.Writer) error {
+	// Marshalling part is taken from: https://github.com/r3labs/sse/blob/c6d5381ee3ca63828b321c16baa008fd6c0b4564/http.go#L16
+	if len(ev.Data) == 0 && len(ev.Comment) == 0 {
+		return nil
+	}
+
+	if len(ev.Data) > 0 {
+		if _, err := fmt.Fprintf(w, "id: %s\n", ev.ID); err != nil {
+			return err
+		}
+
+		sd := bytes.Split(ev.Data, []byte("\n"))
+		for i := range sd {
+			if _, err := fmt.Fprintf(w, "data: %s\n", sd[i]); err != nil {
+				return err
+			}
+		}
+
+		if len(ev.Event) > 0 {
+			if _, err := fmt.Fprintf(w, "event: %s\n", ev.Event); err != nil {
+				return err
+			}
+		}
+
+		if len(ev.Retry) > 0 {
+			if _, err := fmt.Fprintf(w, "retry: %s\n", ev.Retry); err != nil {
+				return err
+			}
+		}
+	}
+
+	if len(ev.Comment) > 0 {
+		if _, err := fmt.Fprintf(w, ": %s\n", ev.Comment); err != nil {
+			return err
+		}
+	}
+
+	if _, err := fmt.Fprint(w, "\n"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteEvent writes the given data to the given ResponseWriter as an Event.
+func WriteEvent(w *echo.Response, data any) error {
+	b, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	event := Event{
+		Data: b,
+	}
+
+	if err := event.MarshalTo(w); err != nil {
+		return err
+	}
+
+	w.Flush()
+
+	return nil
+}
--- a/domain/mvc/orderbook.go
+++ b/domain/mvc/orderbook.go
@@ -18,7 +18,7 @@ type OrderBookUsecase interface {
 	GetActiveOrders(ctx context.Context, address string) ([]orderbookdomain.LimitOrder, bool, error)
 
 	// GetActiveOrdersStream returns a channel for streaming limit orderbook orders for a given address.
-	// The caller should range over the channel, but note that channel is never closed since there may by multiple
+	// The caller should range over the channel, but note that channel is never closed since there may be multiple
 	// sender goroutines.
 	GetActiveOrdersStream(ctx context.Context, address string) <-chan orderbookdomain.OrderbookResult
 }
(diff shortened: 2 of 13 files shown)


delivery/http/event.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if _, err := fmt.Fprint(w, "\n"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func WriteEvent(w *echo.Response, data any) error {
```
deividaspetraitis, line 81: Add docs

domain/orderbook/order.go
```diff
@@ -93,3 +93,10 @@ type LimitOrder struct {
 	BaseAsset        Asset        `json:"base_asset"`
 	PlacedTx         *string      `json:"placed_tx,omitempty"`
 }
+
+
+type ActiveOrder struct {
```
deividaspetraitis: Add docs, came up with better naming.

passthrough/delivery/http/passthrough_handler.go
```diff
@@ -61,6 +72,58 @@ func (a *PassthroughHandler) GetPortfolioAssetsByAddress(c echo.Context) error {
...
+		select {
+		case <-c.Request().Context().Done():
+			return c.NoContent(http.StatusOK)
+		case orders, ok := <-ch:
+			if !ok {
+				return c.NoContent(http.StatusOK)
+			}
+
+			if orders.Error != nil {
+				a.Logger.Error("GET "+c.Request().URL.String(), zap.Error(orders.Error))
```
deividaspetraitis, line 106: Do we want to respond with an error here @crnbarr93? 

It would look something like: "id: \ndata: { "message": "error" }\n\n"

orderbook/usecase/orderbook_usecase.go
```diff
@@ -137,6 +137,63 @@ func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.P
 	return nil
 }
 
+// GetActiveOrdersStream implements mvc.OrderBookUsecase.
+func (o *OrderbookUseCaseImpl) GetActiveOrdersStream(ctx context.Context, address string)  <-chan orderbookdomain.ActiveOrder {
```
deividaspetraitis: TODO: cover with tests

p0mvn:
Nice!

Sanity check: are there any tests that it would make sense to add to reach the 50% min coverage threshold so that we keep on improving on coverage generally?

orderbook/usecase/orderbook_usecase.go
```diff
@@ -137,6 +137,63 @@ func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.P
 	return nil
 }
 
+// GetActiveOrdersStream implements mvc.OrderBookUsecase.
+func (o *OrderbookUseCaseImpl) GetActiveOrdersStream(ctx context.Context, address string) <-chan orderbookdomain.ActiveOrder {
+	// Fetch orders duration
+	d := 5 * time.Second
+
+	// Result channel
+	c := make(chan orderbookdomain.ActiveOrder, 50) // buffered channel to avoid blocking
```
p0mvn: Let's move these constants to the top of the file (defined in global space) and also list how these numbers were selected (state arbitrarily if so)

orderbook/usecase/orderbook_usecase.go
```diff
@@ -137,6 +137,63 @@ func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.P
 	return nil
 }
 
+// GetActiveOrdersStream implements mvc.OrderBookUsecase.
+func (o *OrderbookUseCaseImpl) GetActiveOrdersStream(ctx context.Context, address string) <-chan orderbookdomain.ActiveOrder {
```
p0mvn: Let's add some tests for this.

p0mvn (commented):
Nice work! Just some nits and a test request.

I'm seeing Sonar code duplication check also failing - is there anything we can address to make it pass?

Also, what is our current integration plan with the FE?

orderbook/usecase/orderbook_usecase.go
deividaspetraitis: I have covered this method with tests [here](https://github.com/osmosis-labs/sqs/pull/518/files#diff-5fa55bd6410efa835405589edaffbb41ba95f810344564647d2d711e8bc15b67R242)

domain/mvc/orderbook.go
```diff
@@ -16,4 +16,9 @@ type OrderBookUsecase interface {
 
 	// GetOrder returns all active orderbook orders for a given address.
 	GetActiveOrders(ctx context.Context, address string) ([]orderbookdomain.LimitOrder, bool, error)
+
+	// GetActiveOrdersStream returns a channel for streaming limit orderbook orders for a given address.
+	// The caller should range over the channel, but note that channel is never closed since there may by multiple
```
p0mvn:
```suggestion
	// The caller should range over the channel, but note that channel is never closed since there may be multiple
```

p0mvn: Who is then responsible for closing it to prevent leakage?

orderbook/usecase/orderbook_usecase.go
```diff
@@ -137,32 +137,95 @@ func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.P
 	return nil
 }
 
+var (
+	// fetchActiveOrdersEvery is a duration in which orders are pushed to the client periodically
+	// This is an arbitrary number selected to avoid spamming the client
+	fetchActiveOrdersDuration = 5 * time.Second
```
p0mvn: I think that we can even make it higher like 10s

orderbook/usecase/orderbook_usecase.go
```diff
@@ -137,32 +137,95 @@ func (o *OrderbookUseCaseImpl) ProcessPool(ctx context.Context, pool sqsdomain.P
...
+	getActiveOrdersStreamChanLen = 50
+)
+
+// GetActiveOrdersStream implements mvc.OrderBookUsecase.
+func (o *OrderbookUseCaseImpl) GetActiveOrdersStream(ctx context.Context, address string) <-chan orderbookdomain.OrderbookResult {
+	// Result channel
+	c := make(chan orderbookdomain.OrderbookResult, getActiveOrdersStreamChanLen)
+
+	// Function to fetch orders
+	fetchOrders := func(ctx context.Context) {
```
p0mvn, line 156: Is there any way we could reuse this in `GetActiveOrders` to reduce code duplication?

passthrough/delivery/http/passthrough_handler.go
```diff
@@ -26,14 +29,21 @@ func formatPassthroughResource(resource string) string {
...
 		OUsecase: ou,
+		Logger:   logger,
 	}
 
 	e.GET(formatPassthroughResource("/portfolio-assets/:address"), handler.GetPortfolioAssetsByAddress)
 	e.GET(formatPassthroughResource("/active-orders"), handler.GetActiveOrders)
+	e.GET(formatPassthroughResource("/active-orders"), func(c echo.Context) error {
+		if c.QueryParam("sse") != "" {
+			return handler.GetActiveOrdersStream(c) // Server-Sent Events (SSE)
+		}
```
p0mvn, line 44: Is there any consideration that we should make wrt load balancing?

p0mvn (approved):
LGTM

domain/mvc/orderbook.go
deividaspetraitis: Since channels are not like files it's fine to keep those open. Closing is necessary when we want to inform client that there will be no values send,  which is simple to achieve with one sender, but difficult with mulitple senders, because sending to close channel would cause a panic. In our scenario sender will have always some values to sent, thus I think adding synchronization to close a channel would only introduce unecessary complexity without clear benefit.

orderbook/usecase/orderbook_usecase.go
deividaspetraitis: I think atm code is quite opimised it for duplication: we're reusing common data structures; I think further optmisations would add not significant improvements vs time spent finding right abstractions due smalll differences between `fetchOrders` and `GetActiveOrders`. For example in `fetchOrders` we skip empty results, have additional logging specifc for this method.

passthrough/delivery/http/passthrough_handler.go
deividaspetraitis: For a reference: discussed via slack.

PR #523, opened 2024-09-29
observability: format logs as json

p0mvn:
https://linear.app/osmosis/issue/SQS-402/improve-sqs-logs-formatting-for-better-datadog-ingestion

## Summary by CodeRabbit

- **New Features**
	- Enhanced logger configuration based on the environment, with improved output for production mode.
- **Bug Fixes**
	- Improved execution flow during application startup by adjusting the order of operations.
- **Chores**
	- Updated dependency version for `github.com/osmosis-labs/sqs/sqsdomain` to ensure compatibility and access to the latest features.

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -108,11 +108,6 @@ func main() {
 		panic(err)
 	}
 
-	// If fails, it means that the node is not reachable
-	if _, err := chainClient.GetLatestHeight(ctx); err != nil {
-		panic(err)
-	}
-
 	encCfg := app.MakeEncodingConfig()
 
 	// logger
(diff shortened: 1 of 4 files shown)


app/main.go
```diff
@@ -122,6 +117,11 @@ func main() {
 	}
 	logger.Info("Starting sidecar query server")
 
+	// If fails, it means that the node is not reachable
```
deividaspetraitis, line 120: Is this change intentional?

deividaspetraitis (approved):
LGTM! :+1:

app/main.go
p0mvn: Yes, this is so that we can get a log "Starting sidecar query server" without the dependency on the node

PR #524, opened 2024-10-08
BE-586 | Claimbot

deividaspetraitis:
This PR introduces The Orderbook Claimbot plugin. Orderbook Claimbot plugin is a plugin that claims filled order book orders.

It scans all active orders for each order book determining which orders have been filled and need to be claimed. At the moment order is said to be claimable if it is filled 98 percent or more. In order for an order book to be processed to claim its active orders it must be canonical as per SQS definition.

At this moment there is a few drawbacks of current implementation that can be improved in future, such as for example claiming orders from multiple orderbooks within single transaction, or using test keyring alternative.

## Release Notes

- **New Features**
	- Introduced the Orderbook Claim Bot for processing claimable orders.
	- Added new targets for managing the Orderbook Claim Bot in the configuration.
	- New README documentation for the Orderbook Claim Bot detailing setup and usage.
	- Enhanced functionality for batch claim transactions in the Orderbook Claim Bot.
	- Added methods to retrieve claimable orders for specified orderbooks.

- **Bug Fixes**
	- Updated naming conventions for consistency across the Orderbook plugins.

- **Documentation**
	- Enhanced clarity and specificity in the README files for both the Orderbook Claim Bot and Filler Bot.

- **Chores**
	- Improved the backup script for configuration files to prevent overwriting existing backups.
	- Updated the logging functionality to support named loggers.


ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil
+	}
+	defer o.atomicBool.Store(false)
+
+	for _, orderbook := range orderbooks {
+		// TODO: get ticks
+		ticks, ok := o.orderbookRepository.GetAllTicks(orderbook.PoolID)
+		if !ok {
+			// TODO: report an error, this should not happen
+			fmt.Printf("no ticks for orderbook %s\n", orderbook.ContractAddress)
```
p0mvn: let's use logger instead

ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+			continue
+		}
+
+		for _, t := range ticks {
+			// TODO: Do we wont to store all orders inside memory?
+			orders, err := o.orderBookClient.GetOrdersByTick(ctx, orderbook.ContractAddress, t.Tick.TickId)
```
p0mvn: We don't but we could start ingesting orders node side as a future optimization. I think that it is fine to have as is for the timebeing

ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+	// This may be relaxed in the future.
+	if !o.atomicBool.CompareAndSwap(false, true) {
+		o.logger.Info("orderbook claimer is already in progress", zap.Uint64("block_height", blockHeight))
+		return nil
+	}
+	defer o.atomicBool.Store(false)
+
+	for _, orderbook := range orderbooks {
+		// TODO: get ticks
+		ticks, ok := o.orderbookRepository.GetAllTicks(orderbook.PoolID)
```
p0mvn: It might be good to abstract the loop contents in a function so that we can test it

ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+// ProcessEndBlock implements domain.EndBlockProcessPlugin.
+func (o *orderbookClaimerIngestPlugin) ProcessEndBlock(ctx context.Context, blockHeight uint64, metadata domain.BlockPoolMetadata) error {
+	ctx, span := tracer.Start(ctx, "orderbooktFillerIngestPlugin.ProcessEndBlock")
+	defer span.End()
+
+	orderbooks, err := o.poolsUseCase.GetAllCanonicalOrderbookPoolIDs()
+	if err != nil {
+		o.logger.Error("failed to get all canonical orderbook pool IDs", zap.Error(err))
+		return err
+	}
```
p0mvn: The metadata contains the pools that were modified within a block. We should only process the intersection of modified pools from the `metadata` and `orderbooks`.

Otherwise, we would be doing redundant processing.

ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+
+		atomicBool: atomic.Bool{},
+
+		logger: logger,
+	}
+}
+
+// ProcessEndBlock implements domain.EndBlockProcessPlugin.
+func (o *orderbookClaimerIngestPlugin) ProcessEndBlock(ctx context.Context, blockHeight uint64, metadata domain.BlockPoolMetadata) error {
+	ctx, span := tracer.Start(ctx, "orderbooktFillerIngestPlugin.ProcessEndBlock")
```
p0mvn: This needs renaming

ingest/usecase/plugins/orderbookclaimer/ordebook_filler_ingest_plugin.go
```diff
@@ -0,0 +1,166 @@
...
+	// Calculate claimable orders for the tick by iterating over each active order
+	// and checking each orders percent filled
+	var claimable orderbookdomain.Orders
+	for _, order := range orders {
+		result, err := q.orderbookusecase.CreateFormattedLimitOrder(
+			orderbook,
+			order,
+		)
+
+		// TODO: to config?
```
p0mvn: makes sense

ingest/usecase/plugins/orderbookclaimer/tx.go
```diff
@@ -0,0 +1,332 @@
...
+		Msg:      msgBytes,
+		Funds:    sdk.NewCoins(),
+	}
+
+	err = txBuilder.SetMsgs(&msg)
+	if err != nil {
+		return fmt.Errorf("failed to set messages: %w", err)
+	}
+
+	accountSequence, accountNumber := getInitialSequence(context.TODO(), o.keyring.GetAddress().String())
```
p0mvn: let's propagate context here so that we are not breaking the trace

ingest/usecase/plugins/orderbookclaimer/tx.go
```diff
@@ -0,0 +1,332 @@
...
+
+	_, gas, err := o.simulateMsgs(context.TODO(), []sdk.Msg{&msg})
```
p0mvn: ditto for all contexts (need to be propagated)

ingest/usecase/plugins/orderbookclaimer/tx.go
```diff
@@ -0,0 +1,332 @@
...
+
+func getInitialSequence(ctx context.Context, address string) (uint64, uint64) {
+	// Create a connection to the gRPC server.
+	grpcConn, err := grpc.NewClient(
+		RPC, // Or your gRPC server address.
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		// return err
+	}
```
p0mvn: I think that these helpers are already defined somewhere. Let's make sure that we are not duplicating the code and reuse as much as possible

slices/slices.go
```diff
@@ -0,0 +1,16 @@
...
+	for l := 0; l < len(s); l += size {
+		h := l + size
+		if h > len(s) {
+			h = len(s)
+		}
+		result = append(result, s[l:h])
+	}
+
+	return result
+}
```
p0mvn: Let's move this package under `domain` as opposed to polluting the root level

p0mvn (commented):
Looks good - nice work.

Once the solution is cleaned up with tests added, could you please also link a few DataDog trace samples of processing the block as claimbot?

ingest/usecase/plugins/orderbook/claimbot/README.md
```diff
@@ -0,0 +1,69 @@
...
+    "plugins": [
+        {
+            "name": "orderbook-claimbot-plugin",
+            "enabled": true
+        }
+    ]
+},
+```
+
+Configure the key on a test keyring, and set the following environment variables:
```
cryptomatictrader, line 41: Sorry for lack of context here, key is named as `local` but then what is meant by `local.info`

ingest/usecase/plugins/orderbook/claimbot/order.go
cryptomatictrader, line 1: Sorry for lack of context, should we consider any parallelization here? Guess we shouldn't due to business requirements and so I'm just asking.

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+	orderbook domain.CanonicalOrderBooksResult,
+	orderbookRepository orderbookdomain.OrderBookRepository,
+	orderBookClient orderbookgrpcclientdomain.OrderBookClient,
+	orderbookusecase mvc.OrderBookUsecase,
+	logger log.Logger,
+) (orderbookdomain.Orders, error) {
+	ticks, ok := orderbookRepository.GetAllTicks(orderbook.PoolID)
+	if !ok {
+		return nil, fmt.Errorf("no ticks for orderbook")
+	}
```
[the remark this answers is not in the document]
cryptomatictrader: This looks like a good idea to add context for our log messages!

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+	orderbook domain.CanonicalOrderBooksResult,
+	tick orderbookdomain.OrderbookTick,
+	orderBookClient orderbookgrpcclientdomain.OrderBookClient,
+	orderbookusecase mvc.OrderBookUsecase,
+	logger log.Logger,
+) (orderbookdomain.Orders, error) {
+	orders, err := orderBookClient.GetOrdersByTick(ctx, orderbook.ContractAddress, tick.Tick.TickId)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch orderbook orders by tick ID: %w", err)
+	}
```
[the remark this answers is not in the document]
cryptomatictrader: Ditto

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if len(orders) == 0 {
+		return nil, nil
+	}
+
+	askClaimable := getClaimableOrders(orderbook, orders.OrderByDirection("ask"), tick.TickState.AskValues, fillThreshold, orderbookusecase, logger)
+	bidClaimable := getClaimableOrders(orderbook, orders.OrderByDirection("bid"), tick.TickState.BidValues, fillThreshold, orderbookusecase, logger)
+
+	return append(askClaimable, bidClaimable...), nil
```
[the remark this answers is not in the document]
cryptomatictrader: Good point here. Please consider it , and/or double check `getClaimableOrders` will always return non-nil slice

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// isTickFullyFilled checks if a tick is fully filled by comparing its cumulative total value
+// to its effective total amount swapped.
+func isTickFullyFilled(tickValues orderbookdomain.TickValues) bool {
+	if len(tickValues.CumulativeTotalValue) == 0 || len(tickValues.EffectiveTotalAmountSwapped) == 0 {
+		return false // empty values, thus not fully filled
+	}
+	return tickValues.CumulativeTotalValue == tickValues.EffectiveTotalAmountSwapped
+}
```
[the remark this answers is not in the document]
cryptomatictrader: Is `tickValues.CumulativeTotalValue` String type? If so, this looks like a good point.

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+		config:     config,
+		atomicBool: atomic.Bool{},
+	}, nil
+}
+
+// ProcessEndBlock implements domain.EndBlockProcessPlugin.
+// This method is called at the end of each block to process and claim eligible orderbook orders.
+// ProcessEndBlock implements domain.EndBlockProcessPlugin.
+func (o *claimbot) ProcessEndBlock(ctx context.Context, blockHeight uint64, metadata domain.BlockPoolMetadata) error {
+	ctx, span := tracer.Start(ctx, "orderbooktFillerIngestPlugin.ProcessEndBlock")
```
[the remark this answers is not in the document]
cryptomatictrader: This looks to be a valid comment

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+		o.config.Logger,
+	)
+
+	for _, orderbook := range orders {
+		if orderbook.Err != nil {
+			o.config.Logger.Warn(
+				"failed to retrieve claimable orders",
+				zap.String("contract_address", orderbook.Orderbook.ContractAddress),
+				zap.Error(err),
+			)
```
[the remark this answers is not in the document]
cryptomatictrader: Looks to be a valid comment

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	o.config.Logger.Info("processed end block in orderbook claimer ingest plugin", zap.Uint64("block_height", blockHeight))
+
+	return nil
+}
+
+// processOrderbookOrders processes a batch of claimable orders.
+func (o *claimbot) processOrderbookOrders(ctx context.Context, orderbook domain.CanonicalOrderBooksResult, orders orderbookdomain.Orders) error {
+	if len(orders) == 0 {
+		return fmt.Errorf("no claimable orders found for orderbook %s, nothing to process", orderbook.ContractAddress)
+	}
```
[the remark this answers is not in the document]
cryptomatictrader: This looks to be a nice consideration.

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+			chunk,
+		)
+
+		o.config.Logger.Info("claimed orders",
+			zap.String("orderbook contract address", orderbook.ContractAddress),
+			zap.Any("orders", chunk),
+			zap.Any("tx result", txres),
+			zap.Error(err),
+		)
+
```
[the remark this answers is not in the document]
cryptomatictrader: Looks to be a nice consideration.

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+		// Wait for block inclusion with buffer to avoid sequence mismatch
+		time.Sleep(5 * time.Second)
```
cryptomatictrader: Please consider placing the sleep duration at the top of the file as const

ingest/usecase/plugins/orderbook/claimbot/tx.go
```diff
@@ -0,0 +1,99 @@
...
+			Orders: orders,
+		},
+	}
+
+	msgBytes, err := json.Marshal(batchClaim)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal message: %w", err)
+	}
+	return msgBytes, nil
+}
```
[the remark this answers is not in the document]
cryptomatictrader: Looks to be a nice consideration

cryptomatictrader (commented):
Very nice work! Please consider the comments left. Thank you!

ingest/usecase/plugins/orderbook/claimbot/plugin.go
deividaspetraitis: Moved to the top. :+1:

ingest/usecase/plugins/orderbook/claimbot/README.md
deividaspetraitis: It's full name of the key on the disk as per my understanding, if you would list contents you would see something like that:

```
ll ~/.osmosisd/keyring-test/
total 16
drwxr-xr-x 2 root root 4096 Oct 25 13:54 ./
drwxr-xr-x 7 root root 4096 Oct 25 13:40 ../
-rw------- 1 root root  537 Oct 25 13:54 abcd....address
-rw------- 1 root root  735 Oct 25 13:54 local.info
```

ingest/usecase/plugins/orderbook/claimbot/order.go
deividaspetraitis: I think it is possible to apply parallelization for some calls, for example for procssing different orderbooks. Let me keep this conversation open, and let's see what @p0mvn thinks. In case we decide to do that I think we should create separate task for that.

ingest/usecase/plugins/orderbook/claimbot/tx.go
```diff
@@ -0,0 +1,99 @@
...
+	for i, claim := range claims {
+		orders[i] = []int64{claim.TickId, claim.OrderId}
+	}
+
+	batchClaim := struct {
+		BatchClaim struct {
+			Orders [][]int64 `json:"orders"`
+		} `json:"batch_claim"`
+	}{
+		BatchClaim: struct {
```
p0mvn: Let's address code rabbit rec below

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,160 @@
...
+	if !o.atomicBool.CompareAndSwap(false, true) {
+		o.config.Logger.Info("already in progress", zap.Uint64("block_height", blockHeight))
+		return nil
+	}
+	defer o.atomicBool.Store(false)
+
+	orderbooks, err := getOrderbooks(o.config.PoolsUseCase, blockHeight, metadata)
+	if err != nil {
+		return err
+	}
```
p0mvn: Instead of having this as a function and propagating pools use case as an arg, why not make this be a method on the pools use case?

ingest/usecase/plugins/orderbook/claimbot/config.go
```diff
@@ -0,0 +1,62 @@
...
+
+	txtypes "github.com/cosmos/cosmos-sdk/types/tx"
+)
+
+// Config is the configuration for the claimbot plugin
+type Config struct {
+	Keyring             keyring.Keyring
+	PoolsUseCase        mvc.PoolsUsecase
+	OrderbookUsecase    mvc.OrderBookUsecase
+	OrderbookRepository orderbookdomain.OrderBookRepository
```
p0mvn: We should really only need the use case. Usecase encapsulates the repository. So all requests to the repository should be proxied through the use case.

This is an implementation of the MVC design pattern where the repository is the model.

ingest/usecase/plugins/orderbook/claimbot/config.go
```diff
@@ -0,0 +1,62 @@
...
+	OrderBookClient     orderbookgrpcclientdomain.OrderBookClient
+	AccountQueryClient  authtypes.QueryClient
+	TxfeesClient        txfeestypes.QueryClient
```
p0mvn: Consider abstracting all of the chain clients behind a struct

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+
+// processOrderbooksAndGetClaimableOrders processes a list of orderbooks and returns claimable orders for each.
+func processOrderbooksAndGetClaimableOrders(
+	ctx context.Context,
+	fillThreshold osmomath.Dec,
+	orderbooks []domain.CanonicalOrderBooksResult,
+	orderbookRepository orderbookdomain.OrderBookRepository,
+	orderBookClient orderbookgrpcclientdomain.OrderBookClient,
+	orderbookusecase mvc.OrderBookUsecase,
```
p0mvn: Related to my other comment, we should only need to pass orderbook use case and proxy requests to the repository through it

ingest/usecase/plugins/orderbook/claimbot/order.go
p0mvn: Agreed with the direction of creating a separate task. This should be an easy win that allows us to avoid worrying about any processing latencies in the future.

Let's tackle it before completing the feature.

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+type order struct {
+	Orderbook domain.CanonicalOrderBooksResult
+	Orders    orderbookdomain.Orders
+	Err       error
+}
+
+// processOrderbooksAndGetClaimableOrders processes a list of orderbooks and returns claimable orders for each.
+func processOrderbooksAndGetClaimableOrders(
```
p0mvn, line 22: I'm seeing only this function being covered by tests but there are many downstream function of varying branching complexity.

Could we spend some time covering those with tests?

I think it would pay off in the long-term as it would make any possible refactors drastically easier under simpler test coverage

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,160 @@
...
+			o.config.Logger.Info("failed sending tx",
+				zap.String("orderbook contract address", orderbook.ContractAddress),
+				zap.Any("orders", chunk),
+				zap.Any("tx result", txres),
+				zap.Error(err),
+			)
+		}
+
+		// Wait for block inclusion with buffer to avoid sequence mismatch
+		time.Sleep(blockInclusionWaitTime)
```
p0mvn: Consider querying the sequence number at the beginning of the block and incrementing it in-memory. This should be fine since we have a lock on block processing. That is, if block X is being processed, block X+1 processing cannot start.

This would drastically speed up processing all orders within a block

ingest/usecase/plugins/orderbook/claimbot/order.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/osmosis-labs/sqs/log"
+
+	"go.uber.org/zap"
+)
+
+type order struct {
+	Orderbook domain.CanonicalOrderBooksResult
+	Orders    orderbookdomain.Orders
+	Err       error
+}
```
p0mvn: This abstraction name is confusing.

This is not an order but rather orderbook with its claimable orders.

I suggest renaming it accordingly as well as the variable identifiers to make the logic more readable.

p0mvn (commented):
Nice work! Left comment - please take a look

ingest/usecase/plugins/orderbook/claimbot/plugin.go
deividaspetraitis: My attempt to address this comment within reasonable timeframe did not produced satifying result. The problem I faced was testing. If we would end up moving this func as a method to usecase end result might look like following:

```
func (p *poolsUseCase) GetBlockCanonicalOrderbookPools(metadata domain.BlockPoolMetadata) ([]domain.CanonicalOrderBooksResult, error) {
	orderbooks, err := p.GetAllCanonicalOrderbookPoolIDs()
	if err != nil {
		return nil, err
	}

	var result []domain.CanonicalOrderBooksResult
	for _, orderbook := range orderbooks {
		if _, ok := metadata.PoolIDs[orderbook.PoolID]; ok {
			result = append(result, orderbook)
		}
	}
	return result, nil
}
```

But now, we need to either mock `p.GetAllCanonicalOrderbookPoolIDs` ( which is not easy task ) or setup `Poolsusecase` in a proper way so expected pool IDs are returned in order to cover all scenarios as defined in the [tests](https://github.com/osmosis-labs/sqs/pull/524/files#diff-c34124bfc79d562d366617d7fea859dba1f269644011ff50b798c888272aab10R13). However, that would require a lot of boilerplate as for example [here](https://github.com/osmosis-labs/sqs/blob/BE-586/pools/usecase/pools_usecase_test.go) and we still would not be able to test error case. I think this case is a good example when current design starts limitting us and presents a good opportunity rethink. What do you think?

ingest/usecase/plugins/orderbook/claimbot/order.go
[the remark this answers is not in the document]
deividaspetraitis: Nill slice in `append` is not a problem and expected ( it's not a actual nil value, you can not append `nil` ), you can see example [here](https://go.dev/play/p/vdz2OrwIdXC). Returning nil slice also is not problem, we do return one if there is no orders

ingest/usecase/plugins/orderbook/claimbot/order.go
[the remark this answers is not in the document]
deividaspetraitis: Client of `getClaimableOrdersForTick` passes `orderbook` as agument, tbh I think even current wrapping is a bit redundant thus I have removed it and left for the client to handle the error including adding relevant context if neede.

ingest/usecase/plugins/orderbook/claimbot/order.go
deividaspetraitis: Good catch, updated to `processedOrderbook`.

ingest/usecase/plugins/orderbook/claimbot/order.go
deividaspetraitis: Thanks for the input! Updated `processOrderbooksAndGetClaimableOrders` to process orderbooks to retrieve claimable orders in parallel.

ingest/usecase/plugins/orderbook/claimbot/order.go
deividaspetraitis: Thanks, addressed as per your suggestion!

ingest/usecase/plugins/orderbook/claimbot/config.go
deividaspetraitis: Thanks, addressed as per your suggestion!

domain/cosmos/tx/tx.go
```diff
@@ -119,7 +119,7 @@ func SimulateMsgs(
 	txFactory = txFactory.WithAccountNumber(account.AccountNumber)
 	txFactory = txFactory.WithSequence(account.Sequence)
 	txFactory = txFactory.WithChainID(chainID)
-	txFactory = txFactory.WithGasAdjustment(1.05)
+	txFactory = txFactory.WithGasAdjustment(1.15)
```
cryptomatictrader, line 122: Shall we define a const for this so that it is not like being "buried" somewhere in the code?

ingest/usecase/plugins/orderbook/claimbot/plugin.go
```diff
@@ -0,0 +1,186 @@
...
+
+var _ domain.EndBlockProcessPlugin = &claimbot{}
+
+const (
+	tracerName = "sqs-orderbook-claimbot"
+)
+
+var (
+	tracer        = otel.Tracer(tracerName)
+	fillThreshold = osmomath.MustNewDecFromStr("0.98")
```
cryptomatictrader, line 38: Dito. Consider having a const representing 0.98

cryptomatictrader (commented):
LGTM. Leave final review/approval for @p0mvn

ingest/usecase/plugins/orderbook/claimbot/order.go
deividaspetraitis: Added a several tests in the use case: https://github.com/osmosis-labs/sqs/pull/524/files#diff-5fa55bd6410efa835405589edaffbb41ba95f810344564647d2d711e8bc15b67

deividaspetraitis:
SonarCloud is complaining regarding `plugin.go`

ingest/usecase/plugins/orderbook/claimbot/plugin.go
deividaspetraitis: Addressed that [here](https://github.com/osmosis-labs/sqs/pull/524/files#diff-9c91025a0f9e383b452c4f91a803200bdeb323f4511b300b44e5ab6bdf19c628R204) :+1:

ingest/usecase/plugins/orderbook/claimbot/plugin.go
p0mvn: The linked example looks much cleaner IMO, and this is how I would expect the API to look like if minimizing code smells

I don't understand what the problem with mocking out `poolsUseCase` was. (the link did not point me to anything specific btw)

Happy to not block this PR on this but a task would be greatly appreciated

ingest/usecase/plugins/orderbook/claimbot/plugin.go
p0mvn: Or even in config

p0mvn (approved):
LGTM - great job!

Sanity checking - have we run this on a DO node for 1-2 days to confirm that there are no sequence number errors?

PR #550, opened 2024-11-04
refactor: decouple base fee fetching as part of quote from simulation

p0mvn:
Several changes:
- Pre-compute base fee by fetching it at the end of ingest via an always-enabled plugin
- Add a query param to fetch base fee as part of quote without having to simulate
- Makes simulation optimistic. If it fails, we propagate the error but also propagate the base fee value regardless of the success of the simulation

Ultimately, this will allow us to compute fees more precisely in a more performant way in the Polaris SQS integration while having gas estimates hardcoded. 

The next big improvement would be to implement swap simulations without requiring bank sends. Documented the proposal in this chain [task](https://linear.app/osmosis/issue/CHAIN-691/develop-simulate-swap-endpoint-for-gas-estimation-without-user-address)

## Testing
- Covered by existing tests (unit and integration)
- Added new
- Added API tests
- Tested manually via swagger

## Release Notes

- **New Features**
	- Introduced a new parameter `appendBaseFee` in the `/router/quote` endpoint to include base fees in the quote response.
	- Added functionality to return the base fee in the `/quote` endpoint regardless of simulation success.
	- Enhanced the `/quote` endpoint to support simulations for "out given in" single routes.
	- New struct `TxFeeInfo` to encapsulate transaction fee information, including base fee and error messages.

- **Bug Fixes**
	- Improved error handling in quote simulation and gas calculation processes.

- **Documentation**
	- Updated API documentation to reflect the new `appendBaseFee` parameter and changes in response structures.

- **Chores**
	- Updated changelog entries for recent enhancements and refactors across the codebase.


ingest/usecase/plugins/basefee/base_fee_end_block_update_plugin.go
```diff
@@ -0,0 +1,41 @@
...
+
+	baseFee, err := tx.CalculateFeePrice(ctx, p.txfeesClient)
+	if err != nil {
+		p.logger.Error("failed to calculate fee price", zap.Error(err))
+	}
+
+	p.routerRepository.SetBaseFee(baseFee)
+
+	return nil
+}
```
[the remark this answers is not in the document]
cryptomatictrader: This seems to be a valid suggestion.

quotesimulator/quote_simulator.go
```diff
@@ -67,16 +62,10 @@ func (q *quoteSimulator) SimulateQuote(ctx context.Context, quote domain.Quote,
...
 	baseAccount, err := q.accountQueryClient.GetAccount(ctx, simulatorAddress)
 	if err != nil {
-		return 0, sdk.Coin{}, err
-	}
-
-	// Price the message
-	gasAdjusted, feeCoin, err := q.msgSimulator.PriceMsgs(ctx, q.txFeesClient, q.encodingConfig.TxConfig, baseAccount, q.chainID, swapMsg)
-	if err != nil {
-		return 0, sdk.Coin{}, err
+		return domain.QuotePriceInfo{Err: err.Error()}
```
cryptomatictrader: Will this return an empty Error struct/string, causing the caller consider it is a valid quote price data to be used? Suggest to have some meaning error message in it.

cryptomatictrader (approved):
Nice work! LGTM. Approved and some comments.

domain/cosmos/tx/msg_simulator.go
```diff
@@ -43,19 +45,19 @@ type MsgSimulator interface {
...
-// NewGasCalculator creates a new GasCalculator instance.
-func NewGasCalculator(clientCtx gogogrpc.ClientConn, calculateGas CalculateGasFn) MsgSimulator {
+// NewMsgSimulator creates a new GasCalculator instance.
+func NewMsgSimulator(clientCtx gogogrpc.ClientConn, calculateGas CalculateGasFn, memoryRouterRepository routerrepo.RouterRepository) MsgSimulator {
 	return &txGasCalulator{
-		clientCtx:    clientCtx,
-		calculateGas: calculateGas,
+		clientCtx:              clientCtx,
+		calculateGas:           calculateGas,
+		memoryRouterRepository: memoryRouterRepository,
```
deividaspetraitis, line 60: NOTE: See first this [comment](https://github.com/osmosis-labs/sqs/pull/550#discussion_r1828915127).

Can we instead of pulling in `routerrepo.RouterRepository` define more focused interface, for example:

```go
type FeeSource interface {
	GetBaseFee() domain.BaseFee
}
```

domain/base_fee.go
```diff
@@ -0,0 +1,9 @@
+package domain
+
+import "github.com/osmosis-labs/osmosis/osmomath"
+
+// BaseFee holds the denom and current base fee
+type BaseFee struct {
```
deividaspetraitis, line 6: Do we really want to have new abstraction? I think `sdk.Coin` already would decribe it quite well:

```go
fee := sdk.Coin{Denom: "osmo", Amount: 999}
```

We can have type alias, but I still personally find it redundant:

```go
Fee = skd.Coin
```

domain/cosmos/tx/msg_simulator.go
```diff
@@ -90,13 +92,13 @@ func (c *txGasCalulator) BuildTx(
 		return nil, err
 	}
 
-	gasAdjusted, feecoin, err := c.PriceMsgs(ctx, txfeesClient, encodingConfig.TxConfig, account, chainID, msg...)
-	if err != nil {
-		return nil, err
+	priceInfo := c.PriceMsgs(ctx, encodingConfig.TxConfig, account, chainID, msg...)
+	if priceInfo.Err != "" {
```
deividaspetraitis, line 96: While `PriceInfo` abstraction might have some benefits, I am not sure why we changing error type to string, it seems like this abstraction encapsulates some client logic that requires this specific format? I think `txGasCalulator` should be client indepdendent abstraction.

domain/cosmos/tx/msg_simulator.go
```diff
@@ -145,23 +147,33 @@ func (c *txGasCalulator) SimulateMsgs(encodingConfig cosmosclient.TxConfig, acco
 }
 
 // PriceMsgs implements MsgSimulator.
-func (c *txGasCalulator) PriceMsgs(ctx context.Context, txfeesClient txfeestypes.QueryClient, encodingConfig cosmosclient.TxConfig, account *authtypes.BaseAccount, chainID string, msg ...sdk.Msg) (uint64, sdk.Coin, error) {
+func (c *txGasCalulator) PriceMsgs(ctx context.Context, encodingConfig cosmosclient.TxConfig, account *authtypes.BaseAccount, chainID string, msg ...sdk.Msg) domain.QuotePriceInfo {
```
deividaspetraitis: `txGasCalculator` is used in other contexts such as claimbot too, should we rename  `domain.QuotePriceInfo` to `domain.TxFeeInfo`, for example?

domain/cosmos/tx/msg_simulator.go
```diff
@@ -145,23 +147,33 @@ func (c *txGasCalulator) SimulateMsgs(encodingConfig cosmosclient.TxConfig, acco
...
+	baseFee := c.memoryRouterRepository.GetBaseFee()
```
deividaspetraitis, line 151: Do we want to keep backwards compatibility with previous implementation where we are able to compute fee by making request to chain instead of using pre-computed fee value? For example, is there any possible scenario where we might want to recompute fee for each simulation?

router/repository/memory_router_repository.go
```diff
@@ -38,19 +41,37 @@ type routerRepo struct {
...
 		candidateRouteSearchData: sync.Map{},
+		baseFeeMx:                sync.RWMutex{},
+		baseFee:                  domain.BaseFee{},
 
 		logger: logger,
 	}
 }
 
+// GetBaseFee implements RouterRepository.
+func (r *routerRepo) GetBaseFee() domain.BaseFee {
```
deividaspetraitis, line 66: Router has already a lot of responsibilities, introducing new one has resulted in ripple effect in the code base, notice how many changes in different places was required to make to implement this base fee idea. Additionally, I think router is not good candidate for maintaining `baseFee`, because many other parts in the system will use/benefit of this `baseFee` and each part now will be required to pull `RouterRepository`. 
I would recommend creating instead `fees` ( or similar package ) and move relevant fee logic there, for example:

```go
// GetBaseFee ...
func GetBaseFee() sdk.Coin {
	r.baseFeeMx.RLock()
	defer r.baseFeeMx.RUnlock()
	return r.baseFee
}

// SetBaseFee ...
func  SetBaseFee(baseFee sdk.Coin) {
	r.baseFeeMx.Lock()
	defer r.baseFeeMx.Unlock()
	r.baseFee = baseFee
}
```

Using this approach clients would not even need to have any dependency and can call `fees.GetBaseFee()`.

deividaspetraitis (approved):
This feature is really exciting, great work! :rocket: I would strongly recommend to consider suggestions and commens I have left to improve maintainability, scalability of the code base. I would be more than happy to jump in and help to implement suggested changes.

ingest/usecase/plugins/basefee/base_fee_end_block_update_plugin.go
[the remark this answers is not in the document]
deividaspetraitis: +1, I think we could make one step further by letting client of `ProcessEndBlock` logging message while actual implementation of `ProcessEndBlock` would be responsible only for constructing and adding required context to the error:

```go
func (p *baseFeeEndBlockUpdatePlugin) ProcessEndBlock(ctx context.Context, blockHeight uint64, metadata domain.BlockPoolMetadata) error {
    baseFee, err := tx.CalculateFeePrice(ctx, p.txfeesClient)
    if err != nil {
        return errors.New("failed to calculate fee price %w", err)
    }

    p.routerRepository.SetBaseFee(baseFee)
    return nil
}
```

quotesimulator/quote_simulator.go
p0mvn: I don't understand the concern, This returns a struct with an error embedded that the caller can check

domain/cosmos/tx/msg_simulator.go
p0mvn: Done

router/repository/memory_router_repository.go
p0mvn: I created a separate interface that the router repository extends per your earlier comment so that tx gas calculator depends on a lighter interface.

IMO it is premature to start fragmenting the repository abstractions.

The base fee is used only by the router functionality. As a result, I chose to place it into the router repository

domain/cosmos/tx/msg_simulator.go
p0mvn: The base fee can only be updated after processing transactions. As a result, refetching it once per-block is appropriate

domain/cosmos/tx/msg_simulator.go
p0mvn: Done

domain/cosmos/tx/msg_simulator.go
p0mvn: It was changed to string so that we can return this to the client as a JSON response.

Duplicating `err error` when a string error message is already present feels unnecessary.

Happy for this to be iterated in a follow-up if you're feeling strongly

domain/base_fee.go
p0mvn: 1. This has a `Dec`
2. A separate abstraction is introduced for flexibility of extension (i.e if we need to add another piece of information, we would not need to break 10 different APIs to propagate this through the call stack. Instead, modify a single struct). That was a primary reason that this separate struct was introduced

p0mvn:
Applied recommendations and explained wherever there was a justification to not pursue a suggestion. Thanks for reviews

PR #603, opened 2025-01-17
BE-675 | InGivenOut APIs for balancer, stableswap, transmuter pools

deividaspetraitis:
This pull request implements InGivenOut APIs for balancer, stableswap, transmuter pools, more specifically:

- CalculateTokenInByTokenOut
- ChargeTakerFeeExactOut

New APIs ( and some existing ) were covered with unit test while all other changes are a ripple effect caused by introducing these new APIs.

## Summary by CodeRabbit

- **New Features**
	- Added support for calculating token input amounts based on token output.
	- Introduced methods for charging taker fees on exact output amounts.
	- Enhanced pool routing with more detailed token denomination handling.

- **Improvements**
	- Updated struct field tags for improved JSON serialization.
	- Expanded routable pool interfaces with new calculation methods.

- **Technical Updates**
	- Modified function signatures across multiple pool implementation types.
	- Added token input denomination tracking in candidate pool structures.

### DIFF
--- a/ingest/types/candidate_route.go
+++ b/ingest/types/candidate_route.go
@@ -4,6 +4,7 @@ package types
 // candidate pool to be used for routing.
 type CandidatePool struct {
 	ID            uint64
+	TokenInDenom  string
 	TokenOutDenom string
 }
 
--- a/router/usecase/pools/pool_factory.go
+++ b/router/usecase/pools/pool_factory.go
@@ -16,7 +16,7 @@ import (
 
 // NewRoutablePool creates a new RoutablePool.
 // Panics if pool is of invalid type or if does not contain tick data when a concentrated pool.
-func NewRoutablePool(pool ingesttypes.PoolI, tokenOutDenom string, takerFee osmomath.Dec, cosmWasmPoolsParams cosmwasmdomain.CosmWasmPoolsParams) (domain.RoutablePool, error) {
+func NewRoutablePool(pool ingesttypes.PoolI, tokenInDenom string, tokenOutDenom string, takerFee osmomath.Dec, cosmWasmPoolsParams cosmwasmdomain.CosmWasmPoolsParams) (domain.RoutablePool, error) {
 	poolType := pool.GetType()
 	chainPool := pool.GetUnderlyingPool()
 	if poolType == poolmanagertypes.Concentrated {
@@ -37,6 +37,7 @@ func NewRoutablePool(pool ingesttypes.PoolI, tokenOutDenom string, takerFee osmo
 		return &routableConcentratedPoolImpl{
 			ChainPool:     concentratedPool,
 			TickModel:     tickModel,
+			TokenInDenom:  tokenInDenom,
 			TokenOutDenom: tokenOutDenom,
 			TakerFee:      takerFee,
 		}, nil
@@ -56,6 +57,7 @@ func NewRoutablePool(pool ingesttypes.PoolI, tokenOutDenom string, takerFee osmo
 
 		return &routableBalancerPoolImpl{
 			ChainPool:     balancerPool,
+			TokenInDenom:  tokenInDenom,
 			TokenOutDenom: tokenOutDenom,
 			TakerFee:      takerFee,
 		}, nil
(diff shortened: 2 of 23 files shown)


ingest/types/candidate_route.go
```diff
@@ -4,6 +4,7 @@ package types
 // candidate pool to be used for routing.
 type CandidatePool struct {
 	ID            uint64
+	TokenInDenom  string
```
deividaspetraitis, line 7: Note: Adding new `TokenInDenom` field at this point led to more elegant and simpler to understand implementation than alternative appoach by simply renaming existing field to be used for either ExactIn or ExactOut.

router/usecase/pools/routable_cw_alloy_transmuter_pool.go
```diff
@@ -196,6 +208,32 @@ func (r *routableAlloyTransmuterPoolImpl) CalcTokenOutAmt(tokenIn sdk.Coin, toke
 	return tokenOutAmount, nil
 }
 
+// Calculate the token out amount based on the normalization factors:
+//
+// token_in_amt / token_in_norm_factor = token_out_amt / token_out_norm_factor
+// token_in_amt = token_out_amt * token_in_norm_factor / token_out_norm_factor
+func (r *routableAlloyTransmuterPoolImpl) CalcTokenInAmt(tokenInDenom string, tokenOut sdk.Coin) (osmomath.BigDec, error) {
```
deividaspetraitis: Will be used by once `CalculateTokenInByTokenOut` is implemented in future PRs.

ingest/types/candidate_route.go
p0mvn: Please add a note that only one must be set at a time

router/usecase/pools/pool_factory.go
```diff
@@ -37,6 +37,7 @@ func NewRoutablePool(pool ingesttypes.PoolI, tokenOutDenom string, takerFee osmo
 		return &routableConcentratedPoolImpl{
 			ChainPool:     concentratedPool,
 			TickModel:     tickModel,
+			TokenInDenom:  tokenInDenom,
 			TokenOutDenom: tokenOutDenom,
```
p0mvn, line 41: This `TokenInDenom` and `TokenOutDenom` makes sense in the context of this PR but I am worried that if we were to look at it from a higher level, it would be confusing and unclear that only one of can be set at a time. Is there a way we could enhance the abstraction to reflect the restriction?

router/usecase/pools/routable_balancer_pool.go
```diff
@@ -35,6 +35,16 @@ func (r *routableBalancerPoolImpl) CalculateTokenOutByTokenIn(ctx context.Contex
 	return tokenOut, nil
 }
 
+// CalculateTokenInByTokenOut implements RoutablePool.
+func (r *routableBalancerPoolImpl) CalculateTokenInByTokenOut(ctx context.Context, tokenOut sdk.Coin) (sdk.Coin, error) {
+	tokenIn, err := r.ChainPool.CalcInAmtGivenOut(sdk.Context{}, sdk.Coins{tokenOut}, r.TokenInDenom, r.GetSpreadFactor())
```
p0mvn: let's propagate context just in case (we should always do this to preserve trace ids and various connection info)

router/usecase/pools/routable_cw_transmuter_pool.go
```diff
@@ -73,6 +73,27 @@ func (r *routableTransmuterPoolImpl) CalculateTokenOutByTokenIn(ctx context.Cont
 	return sdk.Coin{Denom: r.TokenOutDenom, Amount: tokenIn.Amount}, nil
 }
 
+// CalculateTokenInByTokenOut implements domain.RoutablePool.
+func (r *routableTransmuterPoolImpl) CalculateTokenInByTokenOut(ctx context.Context, tokenOut sdk.Coin) (sdk.Coin, error) {
+	poolType := r.GetType()
+
+	// Esnure that the pool is concentrated
```
p0mvn:
```suggestion
	// Ensure that the pool is concentrated
```

router/usecase/pools/routable_cw_transmuter_pool.go
```diff
@@ -73,6 +73,27 @@ func (r *routableTransmuterPoolImpl) CalculateTokenOutByTokenIn(ctx context.Cont
...
+	if poolType != poolmanagertypes.CosmWasm {
+		return sdk.Coin{}, domain.InvalidPoolTypeError{PoolType: int32(poolType)}
+	}
+
+	balances := r.Balances
+
+	// Validate token in balance
+	if err := validateTransmuterBalance(tokenOut.Amount, balances, r.TokenInDenom); err != nil {
```
p0mvn, line 88: Could you please double check that the parameters of this function have been renamed appropriately?

p0mvn (approved):

p0mvn:
Nice work!

router/usecase/pools/pool_factory.go
p0mvn: (Since I don't have a specific recommendation and, if you've already spent meaningful time on thinking through this, please feel free not to block on this comment)

deividaspetraitis: Actually both will be set since when we are requesting a quote we specify denoms for both tokenIn and tokenOut, however depending on swap method we specify amount only for one of those.

router/usecase/pools/routable_cw_transmuter_pool.go
deividaspetraitis: Good catch! :+1:

PR #611, opened 2025-02-09
chore: remove toolchain go1.23.3

p0mvn:
toolchain go1.23.3 is incompatible with a pubsub dependency. As a result, if one cleans their mod cache and redownloads the dependencies, they might start having issues building.

The problem has been observed on Macs. This change has been confirmed to resolved the issue in both cases.

## Summary by CodeRabbit

- **Chores**
	- Refined internal dependency management and streamlined build configuration.
	- These behind-the-scenes updates improve overall stability and maintainability without altering end-user functionality.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/osmosis-labs/sqs
 
 go 1.22.7
 
-toolchain go1.23.3
-
 require (
 	cosmossdk.io/math v1.5.0
 	cosmossdk.io/store v1.1.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/osmosis-labs/sqs
 
 go 1.22.7
 
-toolchain go1.23.3
```
deividaspetraitis, line 5: Would it make sense set this to `go1.23.3` same as our `go` directive so we can instruct CI pipelines to download this toolchain instead of relying on whatever it's available on those systems to make our build and pipelines more reliable and reproducible?

deividaspetraitis (approved):
Left one non-blocking comment/question.

go.mod
p0mvn: Let's see if any CI unreliability is encountered - from problems encountered, this directive feels like an overhead with little/unclear marginal benefit

PR #612, opened 2025-02-10
BE-686 | Implement FindCandidateRoutesInGivenOut

deividaspetraitis:
Implements [FindCandidateRoutesInGivenOut](https://github.com/osmosis-labs/sqs/pull/612/files#diff-30e5a903633768344f979bd4964c740e815bcecb0a58efa80e43a92ba8a2a425R242) method for `CandidateRouteSearcher`. Covers it with [bench](https://github.com/osmosis-labs/sqs/pull/612/files#diff-9abfea7ddba1b7533543f782ae1d417715da8bdc0dae7b365cf0b9bbb129ca70R13) and [unit](https://github.com/osmosis-labs/sqs/pull/612/files#diff-5d3480d0ad07788672d78e07daae90c9ab3dd2b5652f0408d8f724a3efca86aaR231) tests.

Implementation is based on `FindCandidateRoutesOutGivenIn` by inverting resulting denominations. Test cases were adjusted to be more strict and to take into account actual specific pools returned and number of routes for each test case to ensure stability for each swap method.

Bench results:

```
goos: linux
goarch: amd64
pkg: github.com/osmosis-labs/sqs/router/usecase
cpu: AMD Ryzen 7 5700X3D 8-Core Processor
BenchmarkCandidateRouteSearcherFindCandidateRoutesOutGivenIn
BenchmarkCandidateRouteSearcherFindCandidateRoutesOutGivenIn-16            10000            104957 ns/op
BenchmarkCandidateRouteSearcherFindCandidateRoutesInGivenOut
BenchmarkCandidateRouteSearcherFindCandidateRoutesInGivenOut-16             9898            106635 ns/op
```

Additionally refactors some of the method names in RouterUsecase to highlight swap method for the operation which results in all that noise in the PR.

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced token swap capabilities that support distinct pathways based on either fixed input amounts or desired output amounts, leading to improved quote accuracy and routing precision.
  
- **Refactor**
  - Streamlined and clarified the overall routing and quote simulation logic to ensure more consistent and precise swap estimations for a better user experience.


domain/mvc/router.go
```diff
@@ -59,19 +59,19 @@ type SimpleRouterUsecase interface {
 type RouterUsecase interface {
 	SimpleRouterUsecase
 
-	// GetOptimalQuote returns the optimal quote for the given tokenIn and tokenOutDenom.
-	GetOptimalQuote(ctx context.Context, tokenIn sdk.Coin, tokenOutDenom string, opts ...domain.RouterOption) (domain.Quote, error)
+	// GetOptimalQuoteOutGivenIn returns the optimal quote for the given token swap method exact amount in.
```
deividaspetraitis, line 62: Note for reviewer: No actual changes to those methods but renaming since we're preparing for introducing alternative methods for InGivenOut swap methods, for example [GetCustomDirectQuoteInGivenOut](https://github.com/osmosis-labs/sqs/pull/607/files#diff-997a7334a630951c5bf9ae1370bfdbeaff2630af12769a6b0eab4952d85ac460R76). Renaming allows to use consistent names, for example `GetOptimalQuote` became `GetOptimalQuoteOutGivenIn` aligning with existing `GetOptimalQuoteOutGivenIn`.

domain/quote_simulator.go
```diff
@@ -8,10 +8,10 @@ import (
...
 // QuoteSimulator simulates a quote and returns the gas adjusted amount and the fee coin.
 type QuoteSimulator interface {
-	// SimulateQuote simulates a quote and returns the gas adjusted amount and the fee coin.
+	// SimulateQuoteOutGivenIn simulates a quote and returns the gas adjusted amount and the fee coin.
 	// CONTRACT:
 	// - Only direct (non-split) quotes are supported.
 	// Retursn error if:
 	// - Simulator address does not have enough funds to pay for the quote.
-	SimulateQuote(ctx context.Context, quote Quote, slippageToleranceMultiplier osmomath.Dec, simulatorAddress string) TxFeeInfo
+	SimulateQuoteOutGivenIn(ctx context.Context, quote Quote, slippageToleranceMultiplier osmomath.Dec, simulatorAddress string) TxFeeInfo
```
deividaspetraitis, line 16: Note for reiviewer: renaming in preparation introducing `SimulateQuoteInGivenOut`.

domain/router.go
```diff
@@ -38,15 +38,15 @@ type Route interface {
...
+	// PrepareResultPoolsOutGivenIn strips away unnecessary fields
 	// from each pool in the route,
 	// leaving only the data needed by client
 	// Runs the quote logic one final time to compute the effective spot price.
 	// Note that it mutates the route.
 	// Computes the spot price of the route.
 	// Returns the spot price before swap and effective spot price.
 	// The token in is the base token and the token out is the quote token.
-	PrepareResultPools(ctx context.Context, tokenIn sdk.Coin, logger log.Logger) ([]RoutablePool, osmomath.Dec, osmomath.Dec, error)
+	PrepareResultPoolsOutGivenIn(ctx context.Context, tokenIn sdk.Coin, logger log.Logger) ([]RoutablePool, osmomath.Dec, osmomath.Dec, error)
```
deividaspetraitis, line 49: Note for reviewer: Renaming in preparation introducing [PrepareResultPoolsInGivenOut](https://github.com/osmosis-labs/sqs/pull/607/files#diff-46c605488eb1052f5bf203f04f77cc6ec4482afef6265daa23ec324b572bfac4R55)

.github/workflows/build.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone Repository
-        uses: actions/checkout@master
+        uses: actions/checkout@v4
```
deividaspetraitis, line 48: Note for reviewer: fixes build ["failed because it uses a deprecated version of `actions/cache: v4.0.2`." error](https://github.com/osmosis-labs/sqs/actions/runs/13267045121?pr=612)

.github/workflows/build.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           name: code-coverage-report
 #          path: app
       - name: Analyze with SonarCloud
-        uses: sonarsource/sonarcloud-github-action@master
+        uses: sonarsource/sonarqube-scan-action@v4
```
deividaspetraitis, line 57: Note for reviewer: [sonarsource/sonarcloud-github-action@master](https://github.com/SonarSource/sonarcloud-github-action) action seems to be deprecated and [no longer works](https://github.com/osmosis-labs/sqs/actions/runs/13267419144), alternative [sonarsource/sonarqube-scan-action@v4](https://github.com/SonarSource/sonarqube-scan-action) fixed the issue.

router/usecase/candidate_routes_test.go
```diff
@@ -35,26 +35,122 @@ func (s *RouterTestSuite) TestCandidateRouteSearcher_HappyPath() {
...
+				1263: {},
+				1265: {},
+				1279: {},
+				1281: {},
+				1399: {},
+				1400: {},
+				1464: {},
+				1477: {},
+				1895: {},
+			},
```
p0mvn, line 72: I think we should avoid hardcoding specific pool IDs. Not sure what exactly this assertion accomplishes

Otherwise, if we regen the state, the tests will break, requiring us to spend time modifying.

p0mvn (approved):
LGTM pending testing question

router/usecase/candidate_routes_test.go
deividaspetraitis: The goal of using specific pool IDs is to ensure that computed candidate routes includes those expected pools.
When implementing `InGivenOut`  I broke the algorithm and it returned diferent routes than `OutGiveIn` but tests did not catched this case.
I agree it's not ideal, in ideal situation we should be testing against small hardcoded set of routes, but since we're testing against huge state file it produces such drawbacks, but I think it's better than leaving as it is.

PR #613, opened 2025-02-10
BE-687 | Implement estimateAndRankSingleRouteQuote for exact amount out swap

deividaspetraitis:
Implements [estimateAndRankSingleRouteQuoteInGivenOut](https://github.com/osmosis-labs/sqs/pull/613/files#diff-a301556464bbd68ff39fc74188e9023c49bc4b1b525ca995cf6db85041727217R88) API method for exact amount out swap method and as a result [CalculateTokenInByTokenOut](https://github.com/osmosis-labs/sqs/pull/613/files#diff-abc7b7210fc55db67d247135049783d3a5c9ae353589484ae05dda6eff73ce79R149) gets implemented too.

## Summary by CodeRabbit

- **New Features**
  - Introduced new methods to calculate required token inputs for desired outputs and to rank available routes—enhancing route estimation and quote generation.
  - Extended token quotes with extra pricing details for improved transparency.

- **Refactor**
  - Streamlined token amount representations across quoting, routing, and price computations to ensure consistency and precision in output values.


router/usecase/quote_in_given_out.go
```diff
@@ -20,12 +22,87 @@ var (
...
+// GetAmountIn implements Quote.
+func (q *quoteExactAmountOut) GetAmountIn() sdk.Coin {
+	if q.quoteExactAmountIn != nil {
+		return q.quoteExactAmountIn.GetAmountIn()
+	}
+	return sdk.Coin{Amount: q.AmountIn}
+}
+
+// GetAmountOut implements Quote.
+func (q *quoteExactAmountOut) GetAmountOut() sdk.Coin {
```
deividaspetraitis, line 46: Note for reviewer: here we're falling back to current implementation that is based on inverting `q.quoteExactAmountIn`, but our new implementation will use those new methods since `q.quoteExactAmountIn` will be not longer set.

router/usecase/optimized_routes.go
```diff
@@ -82,6 +82,71 @@ func (r *routerUseCaseImpl) estimateAndRankSingleRouteQuoteOutGivenIn(ctx contex
 	return finalQuote, routesWithAmountOut, nil
 }
 
+// Returns best quote as well as all routes sorted by amount in and error if any.
+// CONTRACT: router repository must be set on the router.
+// CONTRACT: pools reporitory must be set on the router
+func (r *routerUseCaseImpl) estimateAndRankSingleRouteQuoteInGivenOut(ctx context.Context, routes []route.RouteImpl, tokenOut sdk.Coin, logger log.Logger) (quote domain.Quote, sortedRoutesByAmtOut []RouteWithOutAmount, err error) { //nolint:unused
```
deividaspetraitis, line 88: Note for reviewer: `nolint:unused` directive is temporary and will be omited in the upcoming PRs

router/usecase/route/route.go
```diff
@@ -145,6 +145,37 @@ func (r *RouteImpl) CalculateTokenOutByTokenIn(ctx context.Context, tokenIn sdk.
 	return tokenOut, nil
 }
 
+// CalculateTokenInByTokenOut implements Route.
+func (r *RouteImpl) CalculateTokenInByTokenOut(ctx context.Context, tokenOut sdk.Coin) (tokenIn sdk.Coin, err error) {
```
deividaspetraitis, line 149: Note for reviewer: this gets tested with `TestEstimateAndRankSingleRouteQuoteInGivenOut` test case

router/usecase/optimized_routes.go
```diff
@@ -82,6 +82,71 @@ func (r *routerUseCaseImpl) estimateAndRankSingleRouteQuoteOutGivenIn(ctx contex
...
+		// This might be helpful in alloyed cases where the pool gets imbalanced and runs out of liquidity.
+		// If the original routes were computed only through the zero liquidity token, they will be recomputed
+		// through another token due to changed order.
+
+		// Note: the zero length check occurred at the start of function.
+		tokenOutDenom := routes[0].GetTokenOutDenom()
+
+		r.candidateRouteCache.Delete(formatCandidateRouteCacheKey(tokenOut.Denom, tokenOutDenom))
+		tokenInOrderOfMagnitude := GetPrecomputeOrderOfMagnitude(tokenOut.Amount)
+		r.rankedRouteCache.Delete(formatRankedRouteCacheKey(tokenOut.Denom, tokenOutDenom, tokenInOrderOfMagnitude))
```
p0mvn, line 129: This cache is shared between "out given in" and "in given out"

Let's avoid updating the cache for "in given out" to avoid accidentally breaking "out given in".

However, we can still read from the shared cache while having "out given in" be the only writer. If we read this cache with "in given out", we should make sure that we invert the token in and token out

router/usecase/quote_in_given_out.go
p0mvn: Let's add a comment about this in the code

p0mvn (approved):

router/usecase/optimized_routes.go
deividaspetraitis: Since `estimateAndRankSingleRouteQuoteInGivenOut` is unused at the moment let me merge and open separate PR for that in order to not block whole PR.

deividaspetraitis: Follow up PR: https://github.com/osmosis-labs/sqs/pull/614

PR #618, opened 2025-02-17
BE-692 | Implement GetCustomDirectQuoteInGivenOut API

deividaspetraitis:
Implements `GetCustomDirectQuoteInGivenOut` API for computing exact amount out quotes.

This is smaller chunk of bigger PR-607 ( #607 ) implementing computing exact amount out quotes.

### DIFF
--- a/router/usecase/router_usecase.go
+++ b/router/usecase/router_usecase.go
@@ -450,7 +450,7 @@ func (r *routerUseCaseImpl) GetCustomDirectQuoteOutGivenIn(ctx context.Context, 
 	}
 
 	// create candidate routes with given token out denom and pool ID.
-	candidateRoutes := r.createCandidateRouteByPoolID(tokenOutDenom, poolID)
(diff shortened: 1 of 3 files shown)


router/usecase/router_usecase.go
```diff
@@ -467,6 +467,40 @@ func (r *routerUseCaseImpl) GetCustomDirectQuoteOutGivenIn(ctx context.Context,
...
+	routes, err := r.poolsUsecase.GetRoutesFromCandidates(candidateRoutes, tokenOut.Denom, tokenInDenom)
+	if err != nil {
+		return nil, err
+	}
+
+	// Compute direct quote
+	bestSingleRouteQuote, _, err := r.estimateAndRankSingleRouteQuoteOutGivenIn(ctx, routes, tokenOut, r.logger)
+	if err != nil {
+		return nil, err
+	}
```
p0mvn, line 499: Could you please add a comment that we can use the "out given in" implementation because this is a single pool route and we are providing "token out" as "token in"?

p0mvn (approved):

PR #624, opened 2025-03-09
BE-696 | Feature request to improve service recovery after restart

deividaspetraitis:
This PR enables to start SQS service without early exit when Node is not available.
New behaviour is controlled with configuration variable to avoid discrepancies for the current setup with a defaults for early exit.

Feature can be tested with following command:

```bash
SQS_SKIP_CHAIN_AVAILABILITY_CHECK=true go run app/*.go
```

Note: we don't need here any additional mechanisms since auto reconnect is handled by gRPC.

## Summary by CodeRabbit

- **New Features**
	- Introduced a configurable option to control the behavior when chain availability issues occur.
- **Bug Fixes**
	- Enhanced error handling across critical processes to log errors gracefully rather than causing abrupt application terminations.
	- Adjusted the error handling flow during startup to improve overall application robustness.


pablin-10 (approved):
LGTM

p0mvn (approved):

PR #633, opened 2025-03-27
BE-707 | Skip processing quote requests when system is not healthy

deividaspetraitis:
This PR introduces a middleware for `/router/quote` endpoint that would return early instead of computing a quote when system haven't received data required for computation from the Node resulting in cached empty results.

## Summary by CodeRabbit

- **New Features**
  - Introduced a blockchain state validation layer that ensures quote requests are processed only when the underlying data is current, helping prevent stale or empty quote results.
  
- **Refactor**
  - Enhanced the quote endpoint’s routing logic to seamlessly integrate the new validation, resulting in more robust handling and clearer error responses when data isn’t ready.

### DIFF
--- a/router/delivery/http/middleware/quote_chain_state_validator.go
+++ b/router/delivery/http/middleware/quote_chain_state_validator.go
@@ -1,0 +1,30 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/osmosis-labs/sqs/domain"
+	"github.com/osmosis-labs/sqs/domain/mvc"
+
+	"github.com/labstack/echo/v4"
+)
+
+// QuoteChainStateValidatorMiddleware is a middleware that checks the chain readiness before processing the quote requests.
+// It ensures that SQS has data to process for the quote requests and will not cache empty or outdated quote data resulting in incorrect quotes.
+// NOTE: This approach is safe as it is not relying on external chain info data but rather is using the data that is already available in the SQS,
+// meaning that if Node would go down, the SQS will be able to return already cached quotes - as it works without the middleware.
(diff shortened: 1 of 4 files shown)


router/delivery/http/middleware/quote_chain_state_validator.go
```diff
@@ -0,0 +1,30 @@
...
+// QuoteChainStateValidatorMiddleware is a middleware that checks the chain readiness before processing the quote requests.
+// It ensures that SQS has data to process for the quote requests and will not cache empty or outdated quote data resulting in incorrect quotes.
+// NOTE: This approach is safe as it is not relying on external chain info data but rather is using the data that is already available in the SQS,
+// meaning that if Node would go down, the SQS will be able to return already cached quotes - as it works without the middleware.
+func QuoteChainStateValidatorMiddleware(chainUsecase mvc.ChainInfoUsecase) func(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			_, err := chainUsecase.GetLatestHeight()
+			if err != nil {
+				return c.JSON(http.StatusInternalServerError, domain.ResponseError{Message: "no candidate routes found"})
```
PaddyMc, line 21: Should we return better errors here? Or even different errors for this and below? Why did you choose this response specifically?

PaddyMc: This is a nit tho, will approve

PaddyMc (approved):
Approved, have 1 comment

router/delivery/http/middleware/quote_chain_state_validator.go
deividaspetraitis: Good point! Created a task for us decribing why we did that and to fix it: https://linear.app/osmosis/issue/BE-710/improve-error-messages-for-pr-633-related-to-be-707

PR #641, opened 2025-04-18
BE-723 | Edge case fix for pool liquidity

deividaspetraitis:
This PR fixes a few edge cases for no routes found case and pool liquidity computation.

**Background:**
In order to compute a price for a token we need to compute a route. To compute a route we need to compute a price of tokens for the pools to compute their liquidity. This cyclic dependency is solved by initially setting a low value for expected minimum pool liquidity to `0`, which is initially `0`  to compute a route for a token and derive initial price that will be used to compute actual liquidity of the pool where this token is present.

**Pool liquidity**
As discussed earlier to compute a price of token or pool liquidity we need to compute a route and to do so we set some value of min capitalization for the pools for the the candidate routes. Once several blocks are processed this value is derived by [GetMinPoolLiquidityCap](https://github.com/osmosis-labs/sqs/blob/decda24c2e92e9a0d16fa05edf9af57587d326a2/router/usecase/router_usecase.go#L231). Pools are received and stored "as it is" by calling  [StorePools](https://github.com/osmosis-labs/sqs/blob/decda24c2e92e9a0d16fa05edf9af57587d326a2/pools/usecase/pools_usecase.go#L576)  where some of those has `0` capitalization received from Node. Then  prices are recalculated including the data behind [GetMinPoolLiquidityCap](https://github.com/osmosis-labs/sqs/blob/decda24c2e92e9a0d16fa05edf9af57587d326a2/router/usecase/router_usecase.go#L231) . Issue arises specifically for pools that receives `0` capitalization from Node because at the moment when computing a route for  price/liquidity it reads data from [GetMinPoolLiquidityCap](https://github.com/osmosis-labs/sqs/blob/decda24c2e92e9a0d16fa05edf9af57587d326a2/router/usecase/router_usecase.go#L231) to derive min capitalization value, but because pool has `0` liquidity and data behind [GetMinPoolLiquidityCap](https://github.com/osmosis-labs/sqs/blob/decda24c2e92e9a0d16fa05edf9af57587d326a2/router/usecase/router_usecase.go#L231) is updated after pool liquidity is computed it will read now outdated value and if that value is `> 0` then not routes will be computed and thus price will be computed as `0` and then pool liquidity will be computed as `0` as well.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pool updates to ensure existing liquidity cap values are preserved when new updates do not provide valid values. This prevents loss of important liquidity cap information during pool modifications.

### DIFF
--- a/pools/usecase/pools_usecase.go
+++ b/pools/usecase/pools_usecase.go
@@ -577,6 +577,14 @@ func (p *poolsUseCase) StorePools(pools []ingesttypes.PoolI) error {
 	for _, pool := range pools {
 		// Store pool
 		poolID := pool.GetId()
+		oldPool, err := p.GetPool(poolID)
+		if err == nil {
+			liquidityCap := pool.GetLiquidityCap()
+			if liquidityCap.IsNil() || liquidityCap.Equal(osmomath.ZeroInt()) {
+				pool.SetLiquidityCap(oldPool.GetLiquidityCap())
+			}
+		}
+
 		p.pools.Store(poolID, pool)
 
 		// If orderbook, update top liquidity pool for base and quote denom if it has higher liquidity capitalization.
--- a/pools/usecase/pools_usecase.go
+++ b/pools/usecase/pools_usecase.go
@@ -575,13 +575,15 @@ func (p *poolsUseCase) GetPools(opts ...domain.PoolsOption) ([]ingesttypes.PoolI
 // StorePools implements mvc.PoolsUsecase.
 func (p *poolsUseCase) StorePools(pools []ingesttypes.PoolI) error {
 	for _, pool := range pools {
-		// Store pool
+		// Set liquidity capitalization for the pool to previous value if it is nil or zero.
+		// This value will be used for candidate route search and will be recomputed later.
 		poolID := pool.GetId()
 		oldPool, err := p.GetPool(poolID)
 		if err == nil {
 			liquidityCap := pool.GetLiquidityCap()
 			if liquidityCap.IsNil() || liquidityCap.Equal(osmomath.ZeroInt()) {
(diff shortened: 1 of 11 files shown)


router/repository/memory_router_repository.go
```diff
@@ -167,7 +173,7 @@ func (r *routerRepo) SetCandidateRouteSearchData(data map[string]*domain.Candida
 	r.mu.Lock() // for writers
 	defer r.mu.Unlock()
 
-	oldData, ok := r.candidateRouteSearchData.Load().(candidateRoutes)
+	oldData, ok := r.candidateRouteSearchData.Load().(candidateRoutes) // oldData["factory/osmo10pk4crey8fpdyqd62rsau0y02e3rk055w5u005ah6ly7k849k5tsf72x40/alloyed/allDOGE"]
```
deividaspetraitis: Cleanup.

pools/usecase/pools_usecase.go
```diff
@@ -577,6 +577,14 @@ func (p *poolsUseCase) StorePools(pools []ingesttypes.PoolI) error {
 	for _, pool := range pools {
 		// Store pool
 		poolID := pool.GetId()
+		oldPool, err := p.GetPool(poolID)
+		if err == nil {
```
deividaspetraitis, line 582: Add explanation comments

router/repository/memory_router_repository.go
```diff
@@ -177,7 +183,30 @@ func (r *routerRepo) SetCandidateRouteSearchData(data map[string]*domain.Candida
 
 	maps.Copy(newData, oldData)
 
+	for denom, value := range newData {
```
deividaspetraitis: Add explanation comments

PR #673, opened 2025-07-08
BE-768 | Exclude orderbooks from global token price computation

deividaspetraitis:
This PR excludes orderbook pools from then routes used to derive global token price.

Unfortunately a ripple effect makes this PR bigger than is should be. To make reviews easier below are key parts of this PR:

- Actual fix excluding orderbook pools is [here](https://github.com/osmosis-labs/sqs/pull/673/files#diff-41307ead4f927699fcc0a66d5a154883fc2fab420884eee5018287b94a6fb1bbR146) with a tests covering this edge case [here](https://github.com/osmosis-labs/sqs/pull/673/files)
- To reduce time spent manually debugging pricing related issues from now on pricing endpoint supports debug mode where it returns routes along the the price used to compute the price for the token.
   - HTP level implementation is [here](https://github.com/osmosis-labs/sqs/pull/673/files#diff-2b33546ebb17a93e6624d1da656281d8b50fab06b00084f891647727e9aaa44eR186)
   -  Change to return routes along the price for underlying API is [here](https://github.com/osmosis-labs/sqs/pull/673/files#diff-997a7334a630951c5bf9ae1370bfdbeaff2630af12769a6b0eab4952d85ac460R57)
- Pricing API for HTTP now is covered with unit tests to prevent bugs and breaking changes
  - Hander implemenation is [here](https://github.com/osmosis-labs/sqs/pull/673/files#diff-7188f82c0303ff8f4a1c0b7382a39edf14269c8b8c68bc255d7f87cf59704199)

All other changes are mostly cleanup or ripple effect due the change related with debug mode for prices endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added a debug mode to the `/tokens/prices` API endpoint, allowing users to view detailed route information used in price calculations.
  * API responses for token prices now optionally include route metadata when debug mode is enabled.

* **API Changes**
  * Introduced a new boolean query parameter `debug` to the `/tokens/prices` endpoint.
  * Extended API schemas and documentation to include route details and new fields for price responses.
  * Updated API documentation to reflect new package paths and schema changes.

* **Improvements**
  * Price results now provide both the price and the routes used for calculation, enhancing transparency for users.
  * Increased flexibility for loading mainnet state data in testing utilities.


tokens/delivery/http/tokens_delivery.go
```diff
@@ -201,7 +223,45 @@ func (a *TokensHandler) GetPrices(c echo.Context) (err error) {
...
+			for quoteDenom, price := range quotePrices {
+				result[baseDenom][quoteDenom] = price.Price
+			}
+		}
+		return c.JSON(http.StatusOK, result)
+	}
+
+	// If we are in debug mode, we return the prices with routes
+	result := make(GetPricesDebugResponse, len(prices))
+	for baseDenom, quotePrices := range prices {
```
p0mvn, line 241: Can we add a short README entry about the debug mode please?

Please Encapsulate:
- When we want to use this
-  How to activate this

p0mvn (approved):
LGTM!

tokens/delivery/http/tokens_delivery.go
deividaspetraitis: It will be available under the swagger UI [docs](https://github.com/osmosis-labs/sqs/pull/673/files#diff-2b33546ebb17a93e6624d1da656281d8b50fab06b00084f891647727e9aaa44eR183).
Added a point to include it into our [runbook](https://linear.app/osmosis/issue/BE-754/update-and-pin-sqs-runbook-to-the-channel).

Pls let me know if you still think we should add README section as keeping docs close to the source code should allow us have less outdated docs and make it easier to manage those.

PR #683, opened 2025-10-15
add simple linear retries to Numia APRs query

p0mvn:
Add retries and only alert after 3 failed attempts to reduce the alert noise

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of APR data retrieval with automatic retries and delays on transient failures.
  * Enhanced error reporting and metrics after repeated failures to aid diagnostics.
  * Users should experience fewer intermittent gaps or stale APR values due to network hiccups.
  * No changes required to user settings or workflows; behavior is more resilient behind the scenes.

### DIFF
--- a/sqsutil/datafetchers/numia_apr_fetcher.go
+++ b/sqsutil/datafetchers/numia_apr_fetcher.go
@@ -1,7 +1,9 @@
 package datafetchers
 
 import (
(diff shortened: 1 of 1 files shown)


pablin-10 (approved):
LGTM

PR #686, opened 2025-12-26
fix(router): handle errors in dynamic split route calculations

jasbanza:
Previously, errors returned by CalculateTokenOutByTokenIn in the dynamic programming split optimization were silently ignored with a blank identifier. This caused routes with insufficient liquidity (e.g., orderbook pools) to be incorrectly included in the optimal split calculation, resulting in catastrophic slippage for large swaps.

This fix explicitly checks for errors and treats routes that error as producing zero output, ensuring they are excluded from the optimal split.

Fixes OSMO-53

Testing:
Existing and new unit tests pass (new: TestGetSplitQuote_RouteErrorHandling)

NB: Needs to be merged to stage to test with live data

## Summary by CodeRabbit

* **Bug Fixes**
  * Routes that error during token-amount calculations are now handled and excluded from optimal split results, improving routing reliability.

* **Chores**
  * CI build workflow simplified: SonarCloud trigger removed and code-coverage artifacts are still archived without external Sonar analysis.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/router/usecase/dynamic_splits.go
+++ b/router/usecase/dynamic_splits.go
@@ -218,7 +218,15 @@ func getComputeAndCacheOutAmountCb(ctx context.Context, totalInAmountDec osmomat
 			return curRouteAmt
 		}
 		// This is the expensive computation that we aim to avoid.
-		curRouteOutAmountIncrement, _ := routes[routeIndex].CalculateTokenOutByTokenIn(ctx, sdk.NewCoin(tokenInDenom, inAmountIncrement))
+		curRouteOutAmountIncrement, err := routes[routeIndex].CalculateTokenOutByTokenIn(ctx, sdk.NewCoin(tokenInDenom, inAmountIncrement))
+
+		// If the route errors (e.g., insufficient liquidity in orderbook pools),
+		// treat this route as producing zero output for this increment.
+		// This ensures routes with liquidity issues are not selected by the DP algorithm.
+		if err != nil {
+			routeOutAmtCache[routeIndex][increment] = zero
+			return zero
+		}
 
 		if curRouteOutAmountIncrement.IsNil() || curRouteOutAmountIncrement.IsZero() {
 			curRouteOutAmountIncrement.Amount = zero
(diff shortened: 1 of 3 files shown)


jasbanza:
@ValarDragon fix as discussed. New unit test with mock pools (one for low liq, one for good liq)
Needs merge to stage for e2e testing

jasbanza:
had to first fix sonarcloud trigger:
https://github.com/osmosis-labs/sqs/pull/687

jasbanza:
@alessandrolomanto builds fail. Sonar complains about expired token here: 
https://github.com/osmosis-labs/sqs/actions/runs/20526299029/job/58970477081#step:4:23

router/usecase/dynamic_splits.go
```diff
@@ -218,7 +218,15 @@ func getComputeAndCacheOutAmountCb(ctx context.Context, totalInAmountDec osmomat
...
-		curRouteOutAmountIncrement, _ := routes[routeIndex].CalculateTokenOutByTokenIn(ctx, sdk.NewCoin(tokenInDenom, inAmountIncrement))
+		curRouteOutAmountIncrement, err := routes[routeIndex].CalculateTokenOutByTokenIn(ctx, sdk.NewCoin(tokenInDenom, inAmountIncrement))
+
+		// If the route errors (e.g., insufficient liquidity in orderbook pools),
+		// treat this route as producing zero output for this increment.
+		// This ensures routes with liquidity issues are not selected by the DP algorithm.
+		if err != nil {
+			routeOutAmtCache[routeIndex][increment] = zero
+			return zero
+		}
```
p0mvn, line 229: Curious if this was validated on live data?

p0mvn (approved):
Approving to unblock and make progress but I am curious if this has been tested to fix the issue against live data?

router/usecase/dynamic_splits.go
jasbanza: @p0mvn I wasn't sure how to test with live data, but now I see I can test with the dev docker locally. I'll reproduce some of the requests the frontend makes against it. I have a checklist for manual checks which I'll go through for various assets (which includes this specific case)

jasbanza: Since your comment I read the docs and realized I can just do this locally myself

